PR #8, opened 2025-05-02
add initializer for RegoValue from Encodable

dev-rice:
Call me lazy, but I didn't want to have to import Foundation or type the one-liner to encode my Encodable as JSON and pass it to the `AST.RegoValue.init(jsonData:)`.

This adds a new initializer that accepts an Encodable type, which I imagine being used frequently for authorization checks using dynamic input.

Do feel free to just call me lazy and force me to pass Data everywhere 😂

### DIFF
--- a/Sources/AST/RegoValue+Codable.swift
+++ b/Sources/AST/RegoValue+Codable.swift
@@ -45,6 +45,11 @@ extension RegoValue: Codable {
         try self.init(from: d)
     }
 
+    // Initialize a RegoValue from a type that can be JSON-encoded
+    public init(encodable: Encodable) throws {
+        try self.init(jsonData: try JSONEncoder().encode(encodable))
+    }
+
     // Decodable initializer
     public init(from decoder: Decoder) throws {
         // Try as an object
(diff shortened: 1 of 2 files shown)


Sources/AST/RegoValue+Codable.swift
```diff
@@ -45,6 +45,11 @@ extension RegoValue: Codable {
         try self.init(from: d)
     }
 
+    // Initialize a RegoValue from a type that can be JSON-encoded
+    public init(encodable: Encodable) throws {
```
dev-rice, line 49: I considered putting this in the `init(from: Any)` and adding a `Encodable` case in that switch, but figured a separate initializer was more clear. Interested to hear your alls' thoughts.

Tests/ASTTests/RegoValueTests.swift
```diff
@@ -74,6 +74,65 @@ func testJsonToRegoValues() throws {
...
+        ]
+    )
+    let val = try AST.RegoValue(encodable: input)
+
+    let expected: RegoValue = [
+        "pets": [
+            [
+                "name": "Mr. Meowgi",
+                "age": 4,
+                // With the default JSONEncoder, nil Optionals are not included in the serialized JSON, so 'sibling' is missing.
```
dev-rice, line 120: Pretty sure this will be equivalent evaluation wise, but I just wanted to call it out since this pet in the `RegoValue` differs from the one used in the above test.

PR #9, opened 2025-05-06
Add more bundle validation during query preparation

dev-rice:
The `OPA.Engine.prepareForEvaluation` will now fail if:
- A bundle contains rego source files, but no IR plan
- None of the loaded bundles contained IR plans

Previously, these would only show themselves during prepared query evaluation, but I figured it was more safe to fail early. At least in the applications I've been using this package for, I create the engine and prepare a query on startup, then evaluate the query as needed. When validation is done only during query evaluation, I might not know I loaded an invalid bundle until the first request (or other trigger for policy evaluation) comes in.


Sources/Rego/IREvaluator.swift
```diff
@@ -10,6 +10,17 @@ internal struct IREvaluator {
...
+            // Require at least one IR plan if there are any Rego source files.
+            if bundle.planFiles.isEmpty && !bundle.regoFiles.isEmpty {
+                throw RegoError(
+                    code: .bundleInitializationError,
+                    message: """
+                        intialization failed for bundle \(bundleName), \
+                        found rego source files, but no plan files
+                        """
+                )
+            }
```
patrick-east, line 22: I'm not 100% sure about this check. It is possible that someone might use a bundle for data and it could include rego unrelated to the intended query coming from another bundle.

My 2c, let's keep the check later to make sure at least one bundle had IR plan(s), but I'd prefer to drop this check.

dev-rice: Gotcha, yeah I can see that adding unwanted friction. Updated to only do the check on >0 IR plans.

Sources/Rego/IREvaluator.swift
```diff
@@ -26,6 +26,9 @@ internal struct IREvaluator {
                 }
             }
         }
+        guard !policies.isEmpty else {
```
shomron, line 29: ❓ should we move this check to the _prepare_ phase? I can imagine a future where we continue to add bundles (asynchronously) after the initializer, but before prepareForEvaluation. What do you think?

dev-rice: Sorry, I think I'm missing some context. In the current state, `prepareForEvaluation` is the only thing creating one of these `IREvaluator`s. I can see the longer vision for dynamically adding bundles in for evaluation before a query is prepared. But, for now, this evaluator looks like the one unit of evaluation, and it's also where we can judge the size of `policies` from. Moving it elsewhere would require a new method here, or changing visibility of that property. 

I'll trust your judgement, but it does seem like a bridge we can cross later 🤷

shomron: Apologies - I was reading the change incorrectly - you are correct and the `IREvaluator` is indeed constructed during `prepareForEvaluation` and expected to be provided with all available bundles at that time 👍🏻

shomron (approved):

PR #10, opened 2025-05-06
builtins: numbers.range and range_step

### LINKED ISSUE #6: builtins: implement numbers.range and range_step
Need implementation of [numbers.range](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-numbers-numbersrange) and [numbers.range_step](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-numbers-numbersrange_step) to be used in policies

### Expected Outcome
* `numbers.range` and `numbers.range_step` can be used in policies
* *numbersrange* tests pass: https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/numbersrange

DFrenkel:
Implement numbers.range and numbers.range_step builtins maintaining compatibility with Go version.

Resolves #6

### DIFF
--- a/Sources/Rego/Builtins/Numbers.swift
+++ b/Sources/Rego/Builtins/Numbers.swift
@@ -1,0 +1,72 @@
+import AST
+import Foundation
+
+extension BuiltinFuncs {
+    static func numbersRange(ctx: BuiltinContext, args: [AST.RegoValue]) async throws -> AST.RegoValue {
+        guard args.count == 2 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        return try generateSequence(args: args, withStep: false)
+    }
+
+    static func numbersRangeStep(ctx: BuiltinContext, args: [AST.RegoValue]) async throws -> AST.RegoValue {
+        guard args.count == 3 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 3)
+        }
+
+        return try generateSequence(args: args, withStep: true)
+    }
+
+    private static func generateSequence(args: [AST.RegoValue], withStep: Bool) throws -> RegoValue {
+        guard case .number(_) = args[0] else {
+            throw BuiltinError.argumentTypeMismatch(arg: "a", got: args[0].typeName, want: "number")
+        }
+
+        guard case .number(_) = args[1] else {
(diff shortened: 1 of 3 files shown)


Sources/Rego/Builtins/Numbers.swift
```diff
@@ -0,0 +1,76 @@
...
+            // NOTE that we are okay with this argument being a float with integer value
+            // numbers.range_step(1.0, 3.0, 1.0) works just fine
+            guard let stepValue = args[2].integerValue else {
+                throw BuiltinError.evalError(msg: "step must be integer number but got floating-point number")
+            }
+
+            guard stepValue > 0 else {
+                // Golang error is different: step must be a positive number above zero.
+                // But positive number is by definition above zero...
+                // so using a different error message.
```
srenatus: https://github.com/open-policy-agent/opa/pull/7882/files This has happened since then 😉

Tests/RegoTests/BuiltinTests/NumbersTests.swift
```diff
@@ -0,0 +1,160 @@
...
+    static var allTests: [BuiltinTests.TestCase] {
+        [
+            BuiltinTests.generateFailureTests(
+                builtinName: "numbers.range", sampleArgs: [1, 1], argIndex: 0, argName: "a",
+                allowedArgTypes: ["number"],
+                generateNumberOfArgsTest: true),
+            BuiltinTests.generateFailureTests(
+                builtinName: "numbers.range", sampleArgs: [1, 1], argIndex: 1, argName: "b",
+                allowedArgTypes: ["number"],
+                generateNumberOfArgsTest: false),
```
srenatus: 💭 So I guess these cover the number of args in some way, and since we're doing this with the first arg (above, and in the argName="a" case below), that's sufficient. Sounds fair to me.

srenatus (dismissed):
Limited swift knowledge ⚠️ but the code change looks fine to me.

Tests/RegoTests/BuiltinTests/NumbersTests.swift
DFrenkel: Yes, there are similar calls below each arg....

srenatus (approved):
LGTM

PR #13, opened 2025-05-06
convert Engine initializer to non throwing

dev-rice:
Bundle validation is largely handled in `OPA.Engine.prepareForEvaluation`, and the initializer itself doesn't have any reason to throw.

If we want to reserve the right to throw from this initializer in the future, we can leave this in, but the `init(bundles:)` should also be updated to throw.

At first glance, this looks like a breaking change, but it actually just causes a warning if you still `try` the init. For example, you'd see a warning like this:
>  warning: no calls to throwing functions occur within 'try' expression

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ import Foundation
 
 let path = "some/local/path"
 let bundlePath = OPA.Engine.BundlePath(name: "policyBundle", url: URL(fileURLWithPath: path))
-var regoEngine = try OPA.Engine(bundlePaths: [bundlePath])
+var regoEngine = OPA.Engine(bundlePaths: [bundlePath])
 
 // Prepare does as much pre-processing as possible to get ready to evaluate queries.
 // This only needs to be done once when loading the engine and after updating it.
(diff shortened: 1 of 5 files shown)


patrick-east (approved):
Thanks Chris! I like the shift to more consistently do validation when preparing versus the constructors.

shomron:
@dev-rice thanks for submitting! I think there may be additional call sites that are affected and are generating errors:

```
$ make build

<snip>

.../open-policy-agent/swift-opa/Sources/CLI/EvalCommand.swift:17:26: warning: no calls to throwing functions occur within 'try' expression
15 |     mutating func run() async throws {
16 |         // Initialize a Rego.Engine initially configured with our bundles from the CLI options.
17 |         var regoEngine = try Rego.OPA.Engine(bundlePaths: self.evalOptions.bundlePaths)
   |                          `- warning: no calls to throwing functions occur within 'try' expression
18 |
19 |         let tracer = tracerForLevel(self.evalOptions.explain)

.../open-policy-agent/swift-opa/Sources/CLI/BenchCommand.swift:19:26: warning: no calls to throwing functions occur within 'try' expression
17 |     mutating func run() async throws {
18 |         // Initialize a Rego.Engine initially configured with our bundles from the CLI options.
19 |         var regoEngine = try Rego.OPA.Engine(bundlePaths: self.evalOptions.bundlePaths)
   |                          `- warning: no calls to throwing functions occur within 'try' expression
20 |
21 |         // Prepare does as much pre-processing as possible to get ready to evaluate queries.
```

Could we get those cleaned up as well? 🪣

shomron (approved):

PR #19, opened 2025-05-06
fail query preparation if multiple bundles are loaded

dev-rice:
I was trying to load multiple bundles into the engine and build a single prepared query, but discovered some issues
(https://github.com/open-policy-agent/swift-opa/issues/14, https://github.com/open-policy-agent/swift-opa/issues/18) and learned it's not yet supported.

Until support for this is added, I figured it'd be best to set expectations and fail `prepareForEvaluation` if multiple bundles were loaded. The new bundles for the test case I added are valid and non-overlapping, they can be turned into a success case later.


patrick-east (dismissed):

philipaconrad:
I'm closing this PR, as we've resolved the underlying issue in #110, which merged today, and went out in release `0.0.3`.



Thanks @dev-rice for your early work here-- it was helpful for getting up to speed when I was first looking into this space. 😄

PR #20, opened 2025-05-06
ensure query can be evaluated during preparation

dev-rice:
Continuing my trend of shoving more validation into `OPA.Engine.prepareForEvaluation`, this method now checks that the provided query can be evaluated. This should help catch misconfigurations earlier in application bootstrapping, without necessarily having to evaluate something first.

I think it's a bit of a deviation from how query preparation works with the Go implementation. And I could see this check becoming much more difficult or impossible to implement as query capabilities expand, so I won't be upset if this is rejected.

There are already some tests checking for this behavior, but they were expecting it to happen on `evaluate`. Now the error pops up during `prepareForEvaluation`, but the tests pass either way.


patrick-east (approved):
> I think it's a bit of a deviation from how query preparation works with the Go implementation. And I could see this check becoming much more difficult or impossible to implement as query capabilities expand

Yeah, things get a little blurry with the different evaluator types. The aim was to keep the top level APIs a little more agnostic of how a query got evaluated (IR interpreter, topdown evaluator, etc). There was a point in time on this where it would just give back "undefined" if a query that wasn't in the plan was specified, which is maybe more in line with the Go `rego` APIs.

That being said, if we _are_ going to give back an error saying it isn't a query that can be supported, returning that in the prepare step makes a lot of sense to me.

Sources/Rego/IREvaluator.swift
```diff
@@ -32,24 +32,42 @@ internal struct IREvaluator {
...
     }
+
+    func findPlanAndPolicy(entrypoint: String) -> (IndexedIRPolicy, Plan)? {
+        // TODO: We're assuming that queries are only ever defined in a single policy... that _should_ hold true.. but who's checkin?
+        guard let policy = policies.first(where: { $0.plans[entrypoint] != nil }) else {
+            return nil
+        }
+
+        // The above guard has already ensured that the policy has a plan for the given entrypoint, so force unwrap is safe.
+        return (policy, policy.plans[entrypoint]!)
```
shomron, line 43: ⛏️ reworking this as a for loop to find and return in-line can eliminate the need for a force unwrap and reasoning about whether it is safe in favor of compile-time checks.

Sources/Rego/IREvaluator.swift
```diff
@@ -32,24 +32,42 @@ internal struct IREvaluator {
     init(policies: [IR.Policy]) {
         self.policies = policies.map { IndexedIRPolicy(policy: $0) }
     }
+
+    func findPlanAndPolicy(entrypoint: String) -> (IndexedIRPolicy, Plan)? {
```
shomron, line 36: Does this change warrant a dedicated test?

dev-rice: It's still covered by `IREvaluatorTests.testInvalidEvaluations` because those test cases both prepare a query, then try to evaluate it. Before, the evaluation would fail, but now the query preparation step fails. I considered splitting out a distinct set of tests around only query preparation to put tighter bounds around where the errors are thrown, but ultimately decided against it since it'd be duplicating most of what's covered in `IREvaluatorTests`.

What do you think?

Sources/Rego/IREvaluator.swift
dev-rice: Alternate impl in 2cbbcea4b5f2e0f72a0b997e06325c9dc42aaebd

PR #21, opened 2025-05-07
make PreparedQuery Sendable

dev-rice:
I'm using this package for incoming request authorization in a few different network services. A pattern I've been trying out is to build a `PreparedQuery` on startup, attach it to my server handler class/struct, and call `evaluate` on it for every request. In the HTTP/gRPC frameworks I've been using, these server handler objects have to be `Sendable`, and the easiest way to do so is to make sure all the properties are `Sendable` too.

Right now, as all the implementations of `Evaluator` and `Store`, and `PreparedQuery` itself are all value types, making them conform to `Sendable` was just a matter of bookkeeping. But, assuming API stability is important, this is making an important change. We're signing up for all future implementations of `Store` to be `Sendable`, and `PreparedQuery` to remain so for all time.

Eventually, especially as use cases arise for patching a `Store` after it's been initialized and prepared for a query, I don't think `InMemoryStore` should stay a value type. But, just for conformance's sake now, I think it's fine to keep it and accept that it adds costs for copying potential large data sets to transport across concurrency domains.


patrick-east (dismissed):
Thanks Chris!

Issue #22, opened 2025-05-07
OPA IR Compliance Test Suite


shomron:
Add a compliance test suite to exercise IR interpreter behavior across language implementations.
This suite will leverage and build upon the comprehensive test cases built into OPA in https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1.

Test cases will be colocated with swift-opa initially, but we will explore options to relocate these for easier sharing across language implementations and test harnesses.

PR #23, opened 2025-05-07
add compliance test suite

shomron:
This change adds a compliance test suite under the new ComplianceSuite package. The new package has its own Makefile and Package.swift, the suite can be run from the root directory using `make test-compliance`.

The data files used to drive the test harness are derived from https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1, modified to include generated IR plans and expectations, using a modified `opa-compliance-test` tool which will be published in a separate PR.

Longer term, the test data can be moved and factored in a way that is reusable across all OPA IR implementations to help ensure compatibility.

Closes: #22


patrick-east (approved):
✅

PR #30, opened 2025-09-09
docs: README improvements

anderseknert:
Based on my own noob experience trying to run the code from the examples provided here last night.

- Expand on prerequisite steps without assuming prior knowledge
- Make example complete - i.e. can copy-paste and run
- Build failure due to missing `platforms` definition
- Add community support footer to better engage community

Note that the code in the example now prepares in a separate function, as I thought this better demonstrated the reusability of the returned prepared query. I have no idea whether this is idiomatic or not though, so happy to change if not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,19 +9,67 @@ Rego is a declarative language for expressing policy over structured data. A
 common use of Rego is for defining authorization policy.
 Swift-OPA allows for in-process evaluation of compiled Rego within a Swift-based service or application.
 
+## Prerequisites
+
+### A Rego Policy
+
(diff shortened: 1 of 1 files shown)


shomron (dismissed):
Thanks for the contribution, @anderseknert! This makes things clearer and I believe can help new users adopting the library for the first time!

Minor typo comment below. We can workshop the example (refactor to handle top-level errors with a log instead of panic, etc) - in followups. I also love your idea of adding a micro service example integrating Swift-OPA in the authorization flow!

PR #33, opened 2025-10-09
Expose builtin cache API

philippzagar:
This PR exposes the builtin cache API, allowing **builtins** to access a **shared cache** within policy evaluation via `BuiltinContext/cache`.

Each cache entry is identified by a composite key consisting of a `String`-based key and an optional namespace (defaulting to a global namespace) and maps to a `RegoValue` value.
The cache provides Swift-native access through subscript syntax (`cache[key]`, `cache[key] = value`).

Internally, the cache is backed by a simple Swift dictionary of type `[CompositeKey: RegoValue]`.

### Example

```swift
"custom_cache_builtin": { ctx, args in
    // Read / Write from / to builtin cache
    ctx.cache["test_key"] = .string("test_value")
    let value = ctx.cache["test_key"] ?? .string("fallback")

    return value
}
```

### Usage Notes
- A new cache is automatically created for each policy evaluation via `evaluate()`
- Developers may optionally reuse an existing cache across evaluations (for example, to avoid repeating costly deterministic network requests).


shomron (commented):
Thanks for submitting, @philippzagar!

I have a few comments / questions. Also if you can add some words in the PR description about the change it would be much appreciated! 🙏🏻

anderseknert (commented):
Can't say much about the code, but some questions around docs / communication. Great work, Philip!

PR #35, opened 2025-10-20
fix: benchmark unit conversions and formatting

shomron:
Fix unit conversions and formatting in the CLI `bench` command.


shomron:
NOTE: Adding tests for this has been tricky because the code lives in a CLI target. I'll followup to see if there's a way to make Swift Package Manager happy with that layout. Otherwise we'd need to consider refactoring.

koponen:
> Looks good do me -- is there any implication for #41? Should we try to unify them both at some point...? 🤔

We should merge this patch regardless of the #41 as this is broken as of now. We can clean later.

PR #40, opened 2025-11-04
builtins: tests for verifying consistency

DFrenkel:
Following https://github.com/open-policy-agent/opa/commit/6415ef1, we add equivalent compliance tests to Swift OPA
compliance suite to verify consistent behavior
of `uuid.rfc4122` and `time.now_ns` builtins
when invoked multiple times during a single eval.

### DIFF
--- a/ComplianceSuite/Tests/RegoComplianceTests/TestData/index.json
+++ b/ComplianceSuite/Tests/RegoComplianceTests/TestData/index.json
@@ -5875,6 +5875,12 @@
     ]
   },
   {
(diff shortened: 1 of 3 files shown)


DFrenkel:
> Would it perhaps be a good moment to pull in _all the current tests_ from OPA @ main? We're currently on some old snapshot converted with an internal tool; instead of adding a single file on top of that, I think we should fast-forward our snapshot... 🤔

Yes, agreed. I'll mark this PR as draft and refresh _all_ compliance tests from OPA @ main.

PR #41, opened 2025-12-03
perf: Introduce a benchmark test.

koponen:
Use the benchmark package instead of implementing the benchmark framework here.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -20,6 +20,8 @@ jobs:
       - run: sudo xcode-select -s /Applications/Xcode_26.0.app
       - run: swift --version
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: Install jemalloc
+        run: brew install jemalloc
       - run: make all
       - name: make fmt lint test (ComplianceSuite)
         run: make fmt lint test
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -1,0 +1,30 @@
+// swift-tools-version: 6.0
+
+import PackageDescription
+
+let package = Package(
+    name: "swift-opa-benchmarks",
+    platforms: [
+        .macOS(.v13),
+    ],
+    dependencies: [
+        .package(name: "swift-opa", path: ".."),
+        .package(url: "https://github.com/ordo-one/package-benchmark", from: "1.4.0"),
+    ],
+    targets: [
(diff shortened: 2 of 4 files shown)


srenatus (dismissed):
LGTM. But I cannot really comment on the library choice, except that I've heard good things about it. :+1:

srenatus:
https://github.com/ordo-one/package-benchmark/blob/5dcae6e635a33731a9385be50a972e099b08e429/Sources/Benchmark/Documentation.docc/GettingStarted.md?plain=1#L9 Looks like we need jemalloc

Package.swift
```diff
@@ -24,6 +24,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.5.0"),
         .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0"..<"4.0.0"),
+        .package(url: "https://github.com/ordo-one/package-benchmark", from: "1.4.0"),
```
shomron: 🛑 I don't recall all the details, but we originally used this package and and to back it out because of jemalloc dependency issues, I believe when building in Xcode. I believe the proposed solution to that was to separate the benchmarks into a distinct `Package.swift` under the `Benchmarks` directory, to keep it isolated from the main package.

Related shenanigans: https://github.com/open-policy-agent/swift-opa/commit/45f6574cae88a7c43bc451f19491182ccf519674

shomron (changes requested):

Package.swift
philippzagar: I totally agree, the benchmark should not live within the main `Package.swift` but within the `Benchmarks` folder (depending on the main `Package.swift`) that is totally isolated from the actual library.

philippzagar (changes requested):

Package.swift
koponen: Updated the PR. Note, this required exposing AST, IR, and Rego of the SwiftOPA package.

shomron (changes requested):

Package.swift
```diff
@@ -16,6 +16,15 @@ let package = Package(
...
             targets: ["SwiftOPA"]),
+        .library(
+            name: "AST",
+            targets: ["AST"]),
+        .library(
+            name: "IR",
+            targets: ["IR"]),
+        .library(
+            name: "Rego",
+            targets: ["Rego"]),
```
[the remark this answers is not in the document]
koponen: Fixed the package files.

shomron (approved):

PR #42, opened 2025-12-03
perf: Numeric literals optimization

koponen:
NumberFormatter initialization is expensive. Numeric literals test is ~78% faster, with 67% less heap allocations.

This depends on the earlier benchmark test PR which is included here.


shomron (dismissed):
The actual change looks good - but let's stack it over #41 and remove that code from here.

It would also be interesting to measure the performance of this vs. using a [FormatStyle](https://developer.apple.com/documentation/foundation/formatstyle) approach, but the gains are already compelling.

koponen:
> The actual change looks good - but let's stack it over #41 and remove that code from here.
> 
> It would also be interesting to measure the performance of this vs. using a [FormatStyle](https://developer.apple.com/documentation/foundation/formatstyle) approach, but the gains are already compelling.

That's a good idea. Let's keep this patch simple and then return to FormatStyle once we have a plan for #51. Rebased.

shomron:
Looks like there's a merge conflict to resolve.

koponen:
> Looks like there's a merge conflict to resolve.

Fixed!

srenatus (approved):

PR #43, opened 2025-12-10
Adding string-interpolation support

johanfylling:
By introducing the `internal.template_string` built-in function required for processing template-string processing.

### DIFF
--- a/Sources/Rego/Builtins/Strings.swift
+++ b/Sources/Rego/Builtins/Strings.swift
@@ -479,7 +479,7 @@ func stringifyValue(_ v: RegoValue) throws -> String {
     if case .array(let a) = v {
         return try stringifyArray(a)
     }
-    
+
     if case .set(let s) = v {
         return try stringifySet(s)
     }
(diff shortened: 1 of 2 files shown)


Tests/RegoTests/BuiltinTests/StringsTests.swift
```diff
@@ -1020,6 +1020,76 @@ extension BuiltinTests.StringsTests {
...
+                """)
+        ),
+        BuiltinTests.TestCase(
+            description: "nested set",
+            name: "internal.template_string",
+            args: [.array([
+                .set([.set([.set([])])]),
+            ])],
+            expected: .success("{set()}")
+        ),
```
johanfylling: Some tests are failing right now since `RegoValue` stringification passes through JSON marshaling, which makes e.g. set values not match OPA's string representation. E.g. `[]` and `["a","b"]` instead of `set()` and `{"a", "b"}`.

Sources/Rego/Builtins/Strings.swift
```diff
@@ -447,14 +447,96 @@ extension BuiltinFuncs {
...
+func stringifySet(_ s: Set<RegoValue>) throws -> String {
+    if s.isEmpty {
+        return "set()"
+    }
+
+    var result: String = "{"
+
+    result.append(try s.sorted().map {
+        return try stringifyValue($0)
+    }.joined(separator: ", "))
```
srenatus: ❓ Do we need the space for compatibility? Same for the object.

srenatus (dismissed):
Eventually, we should nuke the extra test data and get to use OPA's latest corpus. But that's not for this PR. Thanks!

srenatus (approved):

PR #45, opened 2025-12-11
builtins: Handle bit shifts overflowing the number representation.

koponen:
Left and right shifts beyond 63 bits trapped.


srenatus (dismissed):

koponen:
> Thanks, @koponen! It would be great if we could add tests covering these cases.

Indeed. Added!

srenatus (approved):

PR #46, opened 2025-12-11
builtins: Handle overflowing of parsed unit values.

koponen:
NSDecimalNumber can express values larger than double. Avoids a trap.


srenatus (approved):
I've done some trivial adjustments to get the lint check to pass.

PR #48, opened 2025-12-12
perf: Reserve capacity for arrays when possible.

koponen:
This is to avoid unnecessary resizing.


srenatus:
It's time we get something merged so you don't need your workflow runs approved ;)

srenatus (approved):

PR #49, opened 2025-12-12
perf: Implement locals as an array instead of dictionary.

koponen:
Locals are allocated consecutively by the planner, making arrays more efficient than dictionaries. Also simplifies scan/with statements by using direct value save/restore instead of scope allocation.

Benchmarking results (using tests in other PRs):

Array Iteration Large (1000 items): 35-36% faster execution, 50% less memory allocation
Array Iteration Medium (100 items): 33-34% faster execution, 47% less memory allocation
Array Iteration Small (10 items): 19-23% faster execution, 32% less memory allocation
Simple Policy Evaluation: 11-17% faster execution, 23% less memory allocation

### DIFF
--- a/Sources/Rego/IREvaluator.swift
+++ b/Sources/Rego/IREvaluator.swift
@@ -138,41 +138,14 @@ internal struct IREvaluationContext {
     }
 }
 
-internal typealias Locals = [IR.Local: AST.RegoValue]
-
 // Frame represents a stack frame used during Rego IR evaluation. Each function call
 // will prompt creation of a new stack frame.
 internal struct Frame {
-    var scopeStack: [Ptr<Scope>] = []
+    var locals: Locals = Locals()
     var results: ResultSet = ResultSet()
 
-    init(locals: Locals = [:]) {
-        _ = self.pushScope(locals: locals)
-    }
-
-    // Create a new Scope, seed with some locals, push it to the stack, and return it.
-    mutating func pushScope(
-        locals: Locals = [:]
-    ) -> Ptr<Scope> {
-        let scopePtr = Ptr(toCopyOf: Scope(locals: locals))
-        self.scopeStack.append(scopePtr)
-        return scopePtr
-    }
-
-    // popScope discards the current scope, and returns its parent scope if there is one.
-    mutating func popScope() -> Ptr<Scope>? {
-        guard self.scopeStack.last != nil else {
-            return nil
-        }
-        self.scopeStack.removeLast()
-        return self.scopeStack.last
-    }
-
-    func currentScope() throws -> Ptr<Scope> {
-        guard !self.scopeStack.isEmpty else {
-            throw RegoError(code: .internalError, message: "no scopes remain on the frame")
-        }
-        return self.scopeStack.last!
+    init(locals: Locals = Locals()) {
+        self.locals = locals
     }
 
     func currentLocation(withContext ctx: IREvaluationContext, stmt: IR.AnyStatement) throws -> OPA.Trace.Location {
@@ -184,26 +157,11 @@ internal struct Frame {
     }
 
     func resolveLocal(idx: IR.Local) -> AST.RegoValue {
-        // walk the scope stack backwards looking for the first hit
-        // stride sequence is empty if we gave an impossible range (-1..0 with -1 stride)
-        // so we correctly don't enter the loop with an empty scopeStack.
-        for i in stride(from: self.scopeStack.count - 1, through: 0, by: -1) {
-            if let localValue = self.scopeStack[i].v.locals[idx] {
-                return localValue
-            }
-        }
-
-        return .undefined
+        return self.locals[idx] ?? .undefined
     }
 
-    mutating func assignLocal(idx: IR.Local, value: AST.RegoValue) throws {
-        guard !self.scopeStack.isEmpty else {
-            throw RegoError(
-                code: .internalError,
-                message: "attempted to assign local with no scope on the frame"
-            )
-        }
-        self.scopeStack.last!.v.locals[idx] = value
+    mutating func assignLocal(idx: IR.Local, value: AST.RegoValue) {
+        self.locals[idx] = value
     }
 
     // TODO, should we throw or return optional on lookup failures?
@@ -311,18 +269,6 @@ internal struct Frame {
     }
 }
 
-// Scopes are children of a Frame, representing a logical Rego scope. Each time
-// a statement has nested blocks a new Scope is pushed. These are primarily used
-// for further protecting locals and tracking "with" value overrides.
-internal struct Scope: Encodable {
-    var locals: Locals
-
-    // Initialize a fresh scope with its own copy of locals
-    init(locals: Locals = [:]) {
-        self.locals = locals
-    }
-}
-
 private struct IRTraceEvent: OPA.Trace.TraceableEvent {
     var operation: OPA.Trace.Operation
     var message: String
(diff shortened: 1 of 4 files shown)


Sources/Rego/IREvaluator.swift
```diff
@@ -314,11 +426,11 @@ internal struct Frame {
 // Scopes are children of a Frame, representing a logical Rego scope. Each time
 // a statement has nested blocks a new Scope is pushed. These are primarily used
 // for further protecting locals and tracking "with" value overrides.
-internal struct Scope: Encodable {
```
srenatus: Does the `Locals()` change make this no longer Encodable? What are the implications, is this acceptable? 🤔

srenatus (dismissed):
I'm unsure about the `Encodable` change, but otherwise this looks good. And a little familiar from the EOPA VM.

Sources/Rego/IREvaluator.swift
```diff
@@ -2,6 +2,119 @@ import AST
...
+    init(_ elements: [AST.RegoValue?]) {
+        self.storage = elements
+    }
+
+    init(repeating value: AST.RegoValue?, count: Int) {
+        self.storage = Array(repeating: value, count: count)
+    }
+
+    init(accommodating locals: [IR.Local], minimumSize: Int = 0) {
+        let maxLocal = locals.max() ?? IR.Local(0)
```
shomron: ❓`Array.max()` is O(N). Will that be appropriate for the main use cases of this constructor?

Sources/Rego/IREvaluator.swift
```diff
@@ -2,6 +2,119 @@ import AST
...
+            return regoValue
+        default:
+            return .undefined
+        }
+    }
+
+    subscript(index: IR.Local) -> AST.RegoValue? {
+        get {
+            let intIndex = Int(index)
+            return intIndex < storage.count ? storage[intIndex] : nil
```
shomron: ⛏️ I know `IR.Local` is unsigned but should we double check the lower bound in case something elsewhere in the code base changes?

```
    guard intIndex >= 0, intIndex < storage.count else {
        return nil
    }
    return storage[intIndex]
```

Sources/Rego/IREvaluator.swift
```diff
@@ -913,14 +1023,12 @@ func evalBlock(
 
             let patched = toPatch.patch(with: overlayValue, at: path)
 
-            // Push a new scope that includes that patched local (likely shadowing a previous
-            // scopes values) and start evaluating the new block
-            var newLocals = currentScopePtr.v.locals
-            newLocals[stmt.local] = patched
-            currentScopePtr = framePtr.v.pushScope(locals: newLocals)
+            // Save original value and set patched value
+            let originalValue = framePtr.v.resolveLocal(idx: stmt.local)
```
shomron: Does `originalValue == toPatch` which we captured a flew lines above?

Sources/Rego/IREvaluator.swift
```diff
@@ -913,14 +1023,12 @@ func evalBlock(
...
+            try framePtr.v.assignLocal(idx: stmt.local, value: patched)
```
shomron: 🛑 If `evalBlock` throws, this will never be patched back to the original.

Tests/RegoTests/IREvaluatorTests.swift
```diff
@@ -211,13 +211,31 @@ struct IRStatementTests {
...
+            self.description = description
+            self.stmt = stmt
+            self.locals = Locals(from: locals)
+            self.funcs = funcs
+            self.staticStrings = staticStrings
+            self.expectLocals = expectLocals.map { Locals(from: $0) }
+            self.ignoreLocals = ignoreLocals
+            self.expectError = expectError
+            self.expectResult = expectResult
+            self.expectUndefined = expectUndefined
```
shomron: I'd suggest removing this in favor of having Locals conform to [ExpressibleByDictionaryLiteral](https://developer.apple.com/documentation/swift/expressiblebydictionaryliteral):

```
extension Locals: ExpressibleByDictionaryLiteral {
    public typealias Key = Int
    public typealias Value = RegoValue

    public init(dictionaryLiteral elements: (Key, Value)...) {
        guard elements.count > 0 else {
            self = Locals()
            return
        }
        
        let maxLocal: Int = elements.reduce(into: 0) { max, elem in
            max = Swift.max(max, elem.0)
        }
        
        var result = Locals(repeating: nil, count: maxLocal + 1)
        
        for (idx, value) in elements {
            result[IR.Local(idx)] = value
        }
        self = result
    }
}
```

Tests/RegoTests/IREvaluatorTests.swift
```diff
@@ -287,7 +305,7 @@ struct IRStatementTests {
                 1: "will be overwritten",
             ],
             expectLocals: [
-                1: .undefined
+                1: AST.RegoValue.undefined
```
shomron: None of these changes are needed with the suggestion above.

Sources/Rego/IREvaluator.swift
```diff
@@ -2,6 +2,119 @@ import AST
 import Foundation
 import IR
 
+// Custom array type that supports IR.Local indexing
+internal struct Locals: Equatable, Sendable, Encodable {
```
shomron: ⛏️ I'd suggest moving this type lower down in the file or to its own `Locals.swift` file. It's a bit out of context before introducing the `IREvaluator` and other related types.

shomron (changes requested):
Thank you, @koponen - this is an exciting improvement! We waffled back and forth on the locals representations when building SwiftOPA - the dictionary approach was a hedge against some underspecified parts of the IR spec + planner. For this new approach to be safe, we should clearly document our assumptions about local allocation, and potentially add validations during our `prepare` phase if any of those are verifiable.

Additionally, we talked about a pre-scan of the IR during the _prepare_ phase to discover the max index, probably hanging on `IndexedIRPolicy`, which we could then use to preallocate the whole backing storage in one shot. Doesn't have to happen in this PR, and it would be interesting to benchmark that particular change.

It also appears that this change removes the needs for scopes on our frames entirely? If that is true, there is additional vestigial support for those that can be removed.

Sources/Rego/IREvaluator.swift
```diff
@@ -2,6 +2,119 @@ import AST
...
+        }
+
+        var result = Locals(repeating: nil, count: max(self.count, rhs.count))
+
+        for i in 0..<self.count {
+            result[IR.Local(i)] = self[IR.Local(i)]
+        }
+
+        for i in 0..<rhs.count {
+            if let rhsValue = rhs[IR.Local(i)] {
```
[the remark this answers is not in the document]
koponen: Improved the comment.

Sources/Rego/IREvaluator.swift
```diff
@@ -2,6 +2,119 @@ import AST
...
+                result[IR.Local(i)] = rhsValue
+            }
+        }
+
+        return result
+    }
+
+    // Resize array to specific size, truncating or extending with nil as needed
+    mutating func resize(to size: Int) {
+        if size < storage.count {
```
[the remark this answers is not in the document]
koponen: Guard added.

Sources/Rego/IREvaluator.swift
koponen: Good catch! Removed.

shomron (approved):
@koponen super, thanks for addressing the feedback and buttoning this up! I'm excited about these performance improvements! 🚀

PR #50, opened 2025-12-12
perf: Reduce allocations in numbers.range.

koponen:
Avoids the map overhead and allows for preallocating the array.


srenatus (dismissed):
LGTM. I guess we could pull the `var result: [RegoValue] = []` and `return .array(result)` out of the conditional, but it doesn't make a huge difference.

shomron (dismissed):
Thanks, @koponen. This looks fine - any meaningful impact to our performance benchmarks?

srenatus (approved):

PR #51, opened 2025-12-17
Introduce a Swift native RegoNumber

koponen:
This removes the heavy reliance on NSNumber.

Improvements are across all benchmarks, with execution time improvements ranging from 1-35% faster. The biggest win is in dynamic calls (23% faster, 31% less memory), while numeric literals show 18-28% speedups. Minor malloc regressions in numeric literal test; short lived transient allocations, most likely due to having to revert to full decimals instead of doubles like NSNumber was able to.

Fixes the compliance suite test arithmetic/test-big-int-0001.json: arithmetic/big_int.


Sources/Rego/IREvaluator.swift
```diff
@@ -721,17 +721,20 @@ func evalBlock(
...
             try framePtr.v.assignLocal(
-                idx: stmt.target, value: .number(NSNumber(value: stmt.value)))
+                idx: stmt.target, value: .number(RegoNumber.int(Int64(stmt.value))))
 
         case .makeNumberRefStmt(let stmt):
             let sourceStringValue = try framePtr.v.resolveStaticString(
                 ctx: ctx, Int(stmt.index))
-            let formatter = NumberFormatter()
-            formatter.numberStyle = .decimal
-            guard let n = formatter.number(from: sourceStringValue) else {
```
srenatus: 🤞 Might fix https://github.com/open-policy-agent/swift-opa/issues/34

Sources/AST/RegoNumber.swift
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /// Get the decimal value
+    public var decimalValue: Decimal {
+        switch self {
+        case .int(let v):
+            // Optimize for common small integer values to avoid repeated Decimal creation
+            if v >= 0 && v <= 1000 {
+                return Self.cachedDecimals[Int(v)]
+            }
```
shomron: 1. Neat, but have we measured to see if this optimization is useful?
2. The range check should be based on `Self.cachedDecimals` bounds, not free-form.

Sources/AST/RegoValue+Codable.swift
```diff
@@ -80,12 +89,13 @@ extension RegoValue: Codable {
                 self = .string(stringValue)
                 return
             }
-            if let doubleValue = try? container.decode(Double.self) {
-                self = .number(NSNumber(value: doubleValue))
+
+            if let intValue = try? container.decode(Int.self) {
+                self = .number(RegoNumber.int(Int64(intValue)))
```
shomron: Should this use `init(value:)`? Should we have additional cases for `UInt.Self`, etc?

```
extension RegoNumber {
    /// Initializer for integer types
    public init<T: BinaryInteger>(value: T) {
        if let int64Value = Int64(exactly: value) {
            self = .int(int64Value)
        } else if let uint64Value = value as? UInt64 {
            self = .decimal(Decimal(uint64Value))
        } else {
            fatalError("Unsupported BinaryInteger type: \(T.self) with value: \(value)")
        }
    }
```

Sources/AST/RegoNumber.swift
```diff
@@ -0,0 +1,287 @@
...
+                -Decimal.greatestFiniteMagnitude
+            self = .decimal(infinityDecimal)
+            return
+        }
+
+        // Check if whole number within Int64 range
+        if let int64Value = decimal.safeInt64Value {
+            self = .int(int64Value)
+            return
+        }
```
shomron: In places where we're constructing these `RegoNumbers`, like in `RegoValue.init(from: decoder)` - we already have special logic for integers. Is this redundant, or still useful in other contexts?

shomron (commented):
@koponen sorry this was delayed - still reviewing but wanted to drop some initial observations. I think at a higher meta level we should discuss / capture what the limitations are with a Decimal-based implementation as compared to our compliance goals, and whether the API will be conducive to future changes we might want to make for arbitrary precision support.

shomron (changes requested):
@koponen I think at a higher level, we should model the `RegoNumber` such that it will not cause more breaking API changes down the road. We'll likely need to swap out `Decimal` with a a type that supports arbitrary precision at some point in the future. It might make sense to hide the internal representation behind private members, and only expose conversion initializers / accessors to extract ints and decimals as needed. This will give us flexibility later to alter our implementation. As such, I'd lean towards a solution where `RegoNumber` is a struct with a private `enum` member as needed.

Thoughts?

srenatus:
I've checked, this will resolve https://github.com/open-policy-agent/swift-opa/issues/34. 🎉

Sources/AST/RegoValue+Codable.swift
```diff
@@ -190,3 +200,23 @@ struct AnyKey: CodingKey {
...
+// Private extension for JSON deserialization
+private extension RegoNumber {
+    /// Create RegoNumber from NSNumber (for JSON deserialization - assumes non-boolean NSNumber)
+    init(_ nsNumber: NSNumber) {
+        let nsNumberIsFloatType = CFNumberIsFloatType(nsNumber as CFNumber)
+
+        if !nsNumberIsFloatType {
+            let nsInt64Value = nsNumber.int64Value
+            let decimalValue = nsNumber.decimalValue
+            if decimalValue == Decimal(nsInt64Value) {
```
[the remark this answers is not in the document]
koponen: Good catch, deferring everything to RegoNumber init.

Sources/AST/RegoValue+Codable.swift
koponen: Indeed, added support for UInt and using value init.

Sources/AST/RegoNumber.swift
```diff
@@ -0,0 +1,287 @@
...
+    public var isFloatType: Bool {
+        switch self {
+        case .int(_):
+            return false
+        case .decimal(_):
+            return true
+        }
+    }
+
+}
```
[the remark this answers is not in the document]
koponen: OPA checks whether the number is representable as integer, not how it's stored, as it's using arbitrary precision numbers only. Interestingly, its Int() implementation does check whether the number is representable as Int64 when determining if it's an integer. I didn't check if all uses of Int() can realistically assume the number to fit into int64.

RegoNumbers that are not constructed with enum directly will normalize whole integers to internal integer representation and isFloatType behaves correctly as long as we stay within int64 limits. To protect from misusing .decimal enum construction, I removed all uses of enum construction outside of AST package but I couldn't figure out if there's an easy way to make compiler provide protection. Any ideas?

We should revisit this for integers beyond int64 once we have an arbitrary precision numbers implemented.

Sources/AST/RegoNumber.swift
```diff
@@ -0,0 +1,287 @@
...
+            self = .decimal(Decimal.nan)
+            return
+        }
+        if !decimal.isFinite {
+            let infinityDecimal = decimal > 0 ?
+                Decimal.greatestFiniteMagnitude :
+                -Decimal.greatestFiniteMagnitude
+            self = .decimal(infinityDecimal)
+            return
+        }
```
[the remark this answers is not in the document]
koponen: Fixed!

koponen:
> @koponen I think at a higher level, we should model the `RegoNumber` such that it will not cause more breaking API changes down the road. We'll likely need to swap out `Decimal` with a a type that supports arbitrary precision at some point in the future. It might make sense to hide the internal representation behind private members, and only expose conversion initializers / accessors to extract ints and decimals as needed. This will give us flexibility later to alter our implementation. As such, I'd lean towards a solution where `RegoNumber` is a struct with a private `enum` member as needed.
> 
> Thoughts?

I changed the enums internal to hide the implementation.

shomron:
@koponen my apologies, I have not been able to take a second pass after those changes. If you'd like, we can time box that and assume good to merge if I haven't said otherwise by tomorrow. Thank you for your patience!

shomron (dismissed):
@koponen great job on this change! I like the direction of hiding the internal storage as it gives us flexibility to iterate on the implementation. I'd even go as far as making the storage private and not internal, but we can look at that later. Let's merge this change and keep refining it in separate PRs. A thought that came up was that it might be good to add (marked as expected to fail) tests for known limitations in round tripping and extremely large numbers, so we can track when we do fix them down the line.

PR #53, opened 2025-12-18
perf: Construct dynamic call path without a path string array.

koponen:
This removes an alloc. The added trivial benchmark shows -1 alloc but no major wall clock time improvement.


srenatus (dismissed):
I don't know if appending to the string causes other allocations, but that aside the change looks good. :+1:

srenatus (approved):

PR #59, opened 2026-01-07
perf: Merge the Frame and MemoCache into IREvaluationContext.

koponen:
This is both to reduce heap allocations and simplify the code.

3-12% reduction in allocs and up to 7% faster execution in benchmarks.


Sources/Rego/IREvaluator.swift
```diff
@@ -12,6 +12,18 @@ private let numberFormatter: NumberFormatter = {
     return formatter
 }()
 
+/// InvocationKey is a key for memoizing an IR function call invocation.
+/// Note we capture the arguments as unresolved operands and not resolved values,
+/// as hashing the values was proving extremely expensive. We instead rely on the
+/// invariant the the plan / evaluator will not modify a local after it has been initally set.
```
srenatus:
```suggestion
/// invariant that the plan / evaluator will not modify a local after it has been initally set.
```

srenatus (dismissed):

srenatus (approved):

Issue #29, opened 2025-09-09
update Compliance Test Suite and "known-issues"


srenatus:
- [ ] https://github.com/open-policy-agent/opa/issues/5311 has been resolved, this should no longer be skipped. But it needs an IR update, since the IR for these policies has changed.
- [ ] there are currently compliance test failures that are not recorded in the exceptions
- [ ] the ComplianceSuite cases are a bit outdated

👉 https://github.com/open-policy-agent/swift-opa/blob/main/ComplianceSuite/Tests/RegoComplianceTests/TestData/rego-compliance.config

PR #60, opened 2026-01-07
builtin: add object.union

sspaink:
implements [object.union](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-objectunion)

fixes 11 failing compliance tests

### DIFF
--- a/Sources/Rego/Builtins/Objects.swift
+++ b/Sources/Rego/Builtins/Objects.swift
@@ -72,30 +72,35 @@ extension BuiltinFuncs {
         return .set(Set(x.keys))
     }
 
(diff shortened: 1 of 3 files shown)


srenatus (approved):
Thanks, nicely done! 🎉 

> fixes 11 failing compliance tests

11/11, though? 😉  I really wish it was easier to tell...

srenatus:
So one thing that would be great, I think, would be to pull off #29. Then the "known issues" reflect realities, and could be updated as part of the PR introducing builtins...

sspaink:
What do you mean by "pull off" #29? Do you mean make child issues for each missing builtin?

PR #61, opened 2026-01-07
swift format the project

sspaink:
Running `make all` locally runs `swift format format --parallel --recursive -i .` which updated a handful of files, thought it would be a good idea to commit these changes.


srenatus (dismissed):
:+1: Might cause some hassle for the open PRs, but I figure if they run the same call, the conflicts should be quick to resolve.

srenatus (approved):
Thank you! 🎈

PR #62, opened 2026-01-07
refactor: Replace Statement protocol with Statement enum to eliminate type-erasure.

koponen:
Renames AnyStatement to Statement and updates all statement usage to wrap concrete types in enum cases.

### DIFF
--- a/Sources/IR/IR+StaticAnalysis.swift
+++ b/Sources/IR/IR+StaticAnalysis.swift
@@ -54,7 +54,7 @@ extension Block {
     }
 }
 
-extension AnyStatement {
+extension Statement {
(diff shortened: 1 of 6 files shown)


srenatus (dismissed):
LGTM but I can't say anything about the performance aspects which were driving https://github.com/open-policy-agent/swift-opa/pull/32 and how they may or may not matter here

koponen:
> LGTM but I can't say anything about the performance aspects which were driving #32 and how they may or may not matter here

This change had no performance impact with our current tests. The type hierarchy is simpler now, though.

srenatus (approved):

PR #63, opened 2026-01-07
perf: Replace enumerated() with indices-based iteration

koponen:
Use direct index access instead of enumerated() overhead in hot loops. Micro enough optimization not to result in a visible change in our benchmarks.


srenatus (approved):
I'm on the fence about this one. If it's not having a measurable impact, we might as well keep the slightly more "swifty" variant? 

But if you feel strongly about this, let's go ahead and merge it.

PR #64, opened 2026-01-07
perf: Pre-allocate locals arrays using static analysis

koponen:
Compute max local for all plans and functions after IR decoding to determine exact array size needed, eliminating runtime growth during evaluation.

Benchmarks show 5-20% reduction in allocs, wall clock time reductions are more modest (a few percent).


srenatus (approved):
Hah, good idea. Those locals won't ever change once we're given the IR. Thanks! 👍

PR #66, opened 2026-01-08
benchmarks: Resolve scope compilation failures with preparedQuery

koponen:
It seems like recent linting pass introduced these.


srenatus:
Odd. The linting shouldn't do that. Also, this PR, which undergoes the linter checks, is green. 💭

srenatus (approved):

PR #67, opened 2026-01-09
perf: Pre-parse numeric literals and pre-verify string indices

koponen:
Introduces a generic IR walker to eliminate code duplication across static analysis passes. Uses it to implement:
    
* Number pre-parsing: parse numeric literals during IR loading to eliminate runtime NumberFormatter calls in MakeNumberRefStmt.
    
* String index verification: validate static string indices during IR loading.

In average benchmarks see ~10% performance improvement (w/ up to 25% less allocations) but numeric literal benchmarks is up to 67% faster and 52% less allocations.


srenatus (approved):
LGTM. Do you reckon it might be worthwhile to add a unit test? Or do we assume that the current e2e rego calls that are tested cover this enough?

PR #68, opened 2026-01-09
ir: Run static analysis on IR policies in engine boot

koponen:
Call prepareForExecution() in custom decoder to compute maxLocal and enable pre-sized locals optimization.


srenatus (dismissed):
LGTM. Does this imply that (potential) users of Swift-OPA need to adjust their code, too, for static analysis? Or was the ComplianceTest scaffolding doing something special?

srenatus (approved):
Sounds good to me! Thanks!

PR #70, opened 2026-01-12
rework MemoStack back into a separate type

shomron:
https://github.com/open-policy-agent/swift-opa/pull/59 was a bit aggressive in flattening the MemoStack functionality into the IREvaluationContext. This led to unclear division of responsibility and APIs - for example having the subscript operator of IREvaluationContext dereference the memo stack is surprising.

This commit works part of that back by separating the MemoStack into its previously distinct type. This should not impact the performance gains of #59, as a lot of that was likely due to removal of the Frame functionality.

### DIFF
--- a/Sources/Rego/IREvaluator.swift
+++ b/Sources/Rego/IREvaluator.swift
@@ -12,18 +12,6 @@ private let numberFormatter: NumberFormatter = {
     return formatter
 }()
 
-/// InvocationKey is a key for memoizing an IR function call invocation.
-/// Note we capture the arguments as unresolved operands and not resolved values,
-/// as hashing the values was proving extremely expensive. We instead rely on the
-/// invariant that the plan / evaluator will not modify a local after it has been initally set.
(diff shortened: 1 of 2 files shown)


koponen:
Any difference in benckmarks?

koponen:
> Any difference in benckmarks?

Most differences are within noise but a few notable changes: 
* an extra malloc per evaluation required.
* dynamic calls see -30% regression at p90,
* numeric literals -5% at p90.

philipaconrad:
@shomron @koponen Is this still a change we want to integrate into the IR evaluator? If so, let me know, and I'll see what needs to be done so that we can get this PR merged in.

koponen:
I suggest we drop this for now because the hot spots have changed drastically with the bytecode VM and justifying any additional mallocs is getting harder by the day.

philipaconrad:
Thanks for taking a look at this one @koponen! We'll close this PR for now.

PR #73, opened 2026-01-15
perf: Locals and args pooling

koponen:
Pool size is bounded by max call depth, ensuring allocations scale with recursion depth not total function invocations.

Pooling optimization helps significantly with non-trivial policies (which involve iteration and function calls), but introduces minor overhead with trivial policies where the work is already cheap. Up to 10% faster with iteration heavy evaluations due to most allocations removed and a couple percent slower with non-iterative, trivial workloads (due to minor increase in allocs).


srenatus (approved):
Sounds good to me. Most policies in the real world should be non-trivial. 👍

PR #39, opened 2025-11-04
builtins: rand.intn

DFrenkel:
Resolves #38

PR #74, opened 2026-01-16
builtins: rand.intn

### LINKED ISSUE #38: builtins: implement rand.intn
Need implementation of [rand.intn](https://www.openpolicyagent.org/docs/policy-reference/builtins/numbers#builtin-numbers-randintn) to be used in policies

### Expected Outcome
* `rand.intn(key, n)` can be used in policies
* The result of `rand.intn(key, n)` call is consistent for the duration of a given evaluation provided both arguments are the same
* All *rand* tests pass: https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/rand

srenatus:
Rebased #39 with a lint fix. I figured it's the quickest way to get this in.

Verified that it's working as expected with the value cached:

```
$ swift run swift-opa-cli eval data.p.r --bundle b0 
[1/1] Planning build
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'swift-opa-cli' complete! (0.82s)
[
  {
    "result" : [
      95
    ]
  }
]
```
Where bundle contained this rego:

```rego
package p

r contains rand.intn("foo", 100) if true
r contains rand.intn("foo", 100) if true
r contains rand.intn("foo", 100) if true
```

Resolves #38

### DIFF
--- a/Sources/Rego/Builtins/Rand.swift
+++ b/Sources/Rego/Builtins/Rand.swift
@@ -1,0 +1,43 @@
+import AST
+import Foundation
+
+private let randNamespace: String = "rand"
+
+extension BuiltinFuncs {
+    static func numbersRandIntN(ctx: BuiltinContext, args: [AST.RegoValue]) async throws -> AST.RegoValue {
+        guard args.count == 2 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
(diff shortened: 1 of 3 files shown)


srenatus:
Since I've done only trivial things here, I'll merge this. I've reviewed @DFrenkel's change in #39.

PR #75, opened 2026-01-16
builtins: add `strings.count`

srenatus:
I've only added arg-type-mismatch test cases, since the actual functionality is asserted by the compliance tests' JSON cases.


koponen (approved):
LGTM

PR #77, opened 2026-01-16
builtins: add `object.union_n`

srenatus:
Again only adding tests not already covered by the compliance-tests.


philipaconrad (approved):
This looks like a good, straightforward implementation of `object.union_n`. Thanks for contributing the PR!

PR #78, opened 2026-01-16
introduce a capabilities.json for swift-opa

srenatus:
This will help

1. People building bundles for usage with swift-opa: they can now check if their rego is compatible at build-time.
2. Making the "supported builtins" information available to other consumers, like the OPA website. See https://github.com/open-policy-agent/opa/issues/8228


philipaconrad (dismissed):
This looks good to me from an OPA / functionality perspective. I'll defer to more experienced Swift devs for style critiques.

PR #79, opened 2026-01-21
RegoTests/Builtins: Extend `generateFailureTests` for custom builtins

philipaconrad:
## What changed?

This PR extends the `generateFailureTests` infrastructure to allow callers to provide a custom `BuiltinRegistry`. This makes it possible to use the failure test case generation for custom builtin implementations.

Previously, "number of args" and "type of args" test cases for a custom builtin would have to be written out manually, which gets tedious quickly.

This slight change in plumbing also means that custom builtin test suites should look *almost* identical to how we test the default builtins. 🙂


srenatus (approved):
👍 LGTM

PR #82, opened 2026-01-26
ci: Add build matrix for Swift 6.0.3 - 6.2.3

philipaconrad:
## What changed?

This PR adds basic support for building and testing the project against each Swift major version (patched) from Swift 6.0, up to Swift 6.2.

This allows us to ensure we support older language versions and toolchains for projects that can't upgrade immediately.

This PR also adds a basic shields.io badge to the README, to indicate our minimum supported version:
[Swift 6.0.3+](https://developer.apple.com/swift/)

## How to test

Like most GH Actions changes, the only way to test if a workflow works is to test it live. ~~This PR will stay in "Draft" state, until I've ensured it actually works as intended.~~

Testing for this PR should be as simple as pushing up a commit and observing the build matrix complete successfully (or not!) for each major language version.

- Success cases:
  - Do we get a green build for each matrix target? (EDIT: ✅ The project builds on all three versions now!)
- Failure cases:
  - Can we introduce a compile issue on a commit and see build failures? (EDIT: ✅ Yep! I reintroduced a 6.0.3-specific issue on a commit, and the 6.0.3 build target exploded.)
  - If the Swift version is wrong, can we detect it and error? (EDIT: ✅ Yep. Added a step that asserts the version is what we think it is.)

## Further Reading
 - `macOS-15` runner image Xcode version support matrix: https://github.com/actions/runner-images/issues/12520
 - `macOS-15` picked up Xcode 26.2: https://github.com/actions/runner-images/pull/13444
 - List of Swift version ↔️ Xcode version mappings: https://swiftversion.net/

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -12,19 +12,51 @@ concurrency:
 
 jobs:
   ci:
-    name: 'make all'
+    name: Build Swift ${{ matrix.swift }}, Xcode ${{ matrix.xcode }}
     runs-on: macos-latest
     permissions:
       contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - xcode: "16.2"
+            swift: "6.0.3"
+          - xcode: "16.4"
+            swift: "6.1" # We can't go more modern than 6.1.0 for the 6.1 runner, due to Xcode version support.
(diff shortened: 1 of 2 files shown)


srenatus (dismissed):
LGTM thank you!

PR #83, opened 2026-01-30
ci: Add Github Release automation.

philipaconrad:
## What changed?

This PR adds a basic Github Actions workflow for creating Github Releases with release notes pulled from the project's `CHANGELOG.md` every time a tag is pushed.

The automation here is based on the OPA project's release setup, where a PR with updated release notes is merged, and then tagged. Once the tag is pushed, the workflow runs that creates a draft Github Release.

Also included is a "release commit check" job from the OPA C# SDK, which helps maintainers make sure they bumped the version numbers consistently for each release. In our case, the only two places we need to check should be just the release commit header and the CHANGELOG.

The release commit check job requires elevated permissions ([`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target)) to be able to take write actions against the host repo, while checking out code in a PR from a fork. In my previous uses of the release-check job, branches were always repo-local, bypassing the security restrictions fork branches have to deal with. The new version of the release-check workflow will not run correctly until it is merged, due to requiring elevated permissions. Once the file exists on `main`, it can run on future PRs.

## How to test?

 - Like most GH Actions PRs, the truest test will be merging and running this PR.
   - We will need a follow-up PR after this one to make sure the release checks work as intended. Logs from earlier Actions runs suggest the scripts are working properly, it's just the PR comment posting that is broken, due to permissions.
 - To test the release notes generation, from the base folder of the project, run the script that extracts the latest release notes: `./tools/build/latest-release-notes.sh --output=RESULT.md`
 - ~~To test the release commit checks, we'll need to push up a fake release commit or two to make sure the scripts are reading from the right places, and that the workflow is behaving correctly.~~

## Further reading

 - OPA's release workflow: https://github.com/open-policy-agent/opa/blob/main/.github/workflows/post-tag.yaml
 - OPA C# SDK's Release PR check system: https://github.com/open-policy-agent/opa-csharp/blob/779663eef8f50650c5182191d517a1ee4e07ca93/.github/workflows/pull-request.yaml#L34-L90

### DIFF
--- a/.github/workflows/post-tag.yaml
+++ b/.github/workflows/post-tag.yaml
@@ -1,0 +1,54 @@
+name: Post Tag
+
+# This workflow contains our release automation.
+on:
+  push:
+    tags:
+      - "*"
+
+permissions:
+  contents: read
+
+jobs:
+  release-build:
+    name: Release Build (macos-latest)
+    runs-on: macos-latest
+    env:
+      DEVELOPER_DIR: "/Applications/Xcode_26.0.app"
+      SWIFT_VERSION: 6.2.3
+    steps:
+      - name: Select Xcode version
+        run: |
+          sudo xcode-select --switch "${DEVELOPER_DIR}/Contents/Developer"
+          xcodebuild -version
+      - name: Show Swift version
+        run: swift --version
+      - name: Check out code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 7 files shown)


srenatus (commented):
Looks good so far! Just one question regarding the tag style, because I don't know the conventions in swift-land.

srenatus (approved):
LGTM! :)

PR #92, opened 2026-02-11
ci: add Linux test runs

srenatus:
This doesn't pass all tests because we haven't sorted out number precision on Linux. For now at least, we can run all the other tests, and the failing ones have been recorded.

Also un-comments an existing test that passes since #51 (probably!)


philipaconrad (approved):
This looks good! Thanks for putting this PR together @srenatus.

Issue #89, opened 2026-02-10
Add support for OPA configs


philipaconrad:
## Description

As part of #57, we'd like to support parsing [OPA config files](https://www.openpolicyagent.org/docs/configuration).

Since YAML is not a supported encode/decode format out-of-box for Swift, we'll focus initially on correctly parsing JSON configs. We may later add support for using the YAML format for configs once the structural bits are plumbed in.

philipaconrad:
I wound up porting over the overwhelming majority of these features in the [Swift OPA SDK](https://github.com/open-policy-agent/swift-opa-sdk), as it was easier to prototype things over there, and we had no restrictions on adding libraries and external dependencies at all.

Now that we've upgraded to Swift 6.1, which introduces [package traits](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/), we may want to look into re-absorbing the config definitions from the Swift OPA SDK.

PR #95, opened 2026-02-17
config: Add initial OPA config structs for `services` top-level key.

philipaconrad:
## What changed?

This commit contains basic struct definitions in Swift for supporting the `services` key in OPA configuration files.

We currently support parsing:
 - Custom plugins
 - Bearer auth
 - Client TLS auth (mTLS)
 - Azure managed identities auth
 - GCP metadata auth

Not supported yet:
 - S3 signing (AWS)
 - OAuth2

The AWS and OAuth2 definitions were left out of this initial implementation, because of how tangled and messy those configs are in the Golang codebase.

## How to test?

 - `make test` picks up the new `Config` module tests automatically.
   - These currently only test that round-tripping serdes works as expected. The test cases were ported over from: [`github.com/open-policy-agent/opa/tree/main/v1/plugins/rest/rest_test.go`](github.com/open-policy-agent/opa/tree/main/v1/plugins/rest/rest_test.go)

## Further reading

 - Partially implements: #89

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -45,6 +45,7 @@ let package = Package(
                 .product(name: "Crypto", package: "swift-crypto", condition: .when(platforms: [.linux])),
             ]
         ),
+        .target(name: "Config"),
         // Internal module tests
         .testTarget(
             name: "ASTTests",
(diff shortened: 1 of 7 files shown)


philipaconrad:
ℹ️ It's looking like the `Config` structs are likely going to land downstream in the [swift-opa-sdk](https://github.com/open-policy-agent/swift-opa-sdk), so I'm closing this PR here to reduce clutter.

PR #97, opened 2026-02-20
AST/RegoValue+Codable: Change decoding order for strings.

philipaconrad:

This PR contains a fix that makes parsing numbers more consistent in formats like YAML where almost any bare number literal can parse as a string, thanks to lax type coercion rules.

By moving string decoding to the last spot in the list of types we try, we get correctly parsed results even in those ambiguous situations.

This came up when attempting to add [`yaml` builtin support over in `swift-opa-sdk`](https://github.com/open-policy-agent/swift-opa-sdk/pull/8). This example parses incorrectly until we change the decoding order:

```yaml
- foo:
  - 1
  - 2
  - 3
```
### Definition of done

 - [x] Change decoding order, and test against the YAML use case in `swift-opa-sdk` manually.
 - [x] Create a regression test case for `swift-opa`.
   - I did this by creating a custom `Decoder` that parses JSON/YAML-style values as we'd expect. For the `init(decoder:)` implementation on `main`, it fails as expected, and after applying the fix on this PR, it passes.

### How to test

This one was uniquely annoying to test, as you need a "permissive" decoder available. I rolled my own for sake of testing.

 - `make test` should still pass, and will pick up the new `RegoValueDecodingTests` suite.
 - After merge, [swift-opa-sdk/8](https://github.com/open-policy-agent/swift-opa-sdk/pull/8) should see its tests go green.

### DIFF
--- a/Sources/AST/RegoValue+Codable.swift
+++ b/Sources/AST/RegoValue+Codable.swift
@@ -88,9 +88,12 @@ extension RegoValue: Codable {
         do {
             let container = try decoder.singleValueContainer()
 
-            if let stringValue = try? container.decode(String.self) {
-                // Normalize string to contiguous UTF-8 for faster comparisons
-                self = .string(String(decoding: stringValue.utf8, as: UTF8.self))
+            if container.decodeNil() {
+                self = .null
+                return
+            }
+            if let boolValue = try? container.decode(Bool.self) {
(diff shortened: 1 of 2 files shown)


srenatus (approved):
LGTM, not a swift expert by any means, but the code change looks straightforward!

PR #99, opened 2026-02-24
ci: Harden CI + Add zizmor static analysis for GH Actions

philipaconrad:
### What changed, and why?

This commit fixes all findings from the Github Actions static analysis tool [`zizmor`](https://docs.zizmor.sh/), and adds it to our pull request checks.

It also changes the release commit detection workflow to use the reduced-permissions `pull_request` workflow trigger, and now skips on non-releases. It succeeds/fails only when a release commit is detected.

Unfortunately, due to the `pull_request_target` already being set up on `main`, *that version* of the workflow is what will run on this PR, so we can't test the new reduced-permissions version until it's merged (breaking the "upstream only" behavior). 🙁 

### Definition of done

 - Do the existing workflows still pass?
 - ~~Do release commits get detected / handled correctly in the detection workflow?~~
   - ~~Skip: Normal PR, no release commit.~~
   - ~~Fail: Release commit, version mismatches.~~
   - ~~Pass: Release commit, versions all match.~~
 - We'll test the release commit detection in a follow-up PR (with the above strategy)

### How to test

 - Allow normal CI to run for this PR.
 - ~~For testing release workflow:~~
   - ~~Skip: tested by the base state of this PR.~~
   - ~~Fail: Push release commit to the PR branch with version mismatch.~~
   - ~~Pass: Push release commit to the PR branch with all versions matched.~~

### Related Resources

 - Prior art: https://github.com/open-policy-agent/opa/pull/8356

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       dependencies:
         patterns:
           - "*"
+    cooldown:
+      default-days: 7
   # Only updates the go.mod when OPA has a version bump.
   - package-ecosystem: "gomod"
     directory: "/tools/generate-compliance-tests"
@@ -16,3 +18,5 @@ updates:
       interval: "weekly"
     allow:
       - dependency-name: "github.com/open-policy-agent/opa"
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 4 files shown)


srenatus (approved):
Also updating the actions! Nice, thanks!

PR #100, opened 2026-02-25
Rego/Store: Add mutex locking around the store's contents.

philipaconrad:

This PR adds basic thread safety for the contents of the Rego Store with a non-recursive mutex lock around the RegoValue holding the store's contents.

Note: This only guards individual read and write ops, and there is no support yet for read or write transactions like we have in upstream OPA.

Known Hazard: Using `Mutex` forces us to upgrade to `macOS >= 15` in our supported platforms. Otherwise, we rapidly descend into a hellscape of trying to shim in a feature backport library, like [`swhitty/swift-mutex`](https://github.com/swhitty/swift-mutex). It can be done, but it's a non-trivial lift for all the conditional compilation stuff. 😓 

### Definition of done

 - [ ] Decide if we want to vendor/include a backport library for `Mutex` support on `macOS < 15` versions (e.g. `swhitty/swift-mutex`)
 - [ ] Test cases for concurrent store writes?

### How to test

 - Any future tests should get picked up in `make test`.

### Related Resources

 - [Synchronization Docs page for `Mutex`](https://developer.apple.com/documentation/synchronization/mutex)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,7 +8,7 @@ import class Foundation.ProcessInfo
 let package = Package(
     name: "swift-opa",
     platforms: [
-        .macOS(.v13),
+        .macOS(.v15),
         .iOS(.v16),
     ],
     products: [
--- a/Sources/Rego/Store.swift
+++ b/Sources/Rego/Store.swift
@@ -1,4 +1,5 @@
 import AST
+import Synchronization
 
 extension OPA {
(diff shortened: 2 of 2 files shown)


philipaconrad:
After talking about it with @shomron, I think we'll keep the more functional/immutable design the Store has used so far, and see how far we can take it before adding mutexes into the mix.

Closing this PR for now-- we can reopen the PR later if we end up needing anything from it.

PR #103, opened 2026-02-26
perf: replace unicode grapheme-cluster-aware string operations with builtins with byte-level equivalents

arirubinstein:

Replaces Unicode grapheme-cluster-aware string operations in the
`contains`, `startsWith`, and `endsWith` builtins with byte-level
equivalents that match Go's strings package semantics.

Why: OPA's builtins mirror Go's strings.Contains(), strings.HasPrefix(),
and strings.HasSuffix(), which all operate on raw bytes. Swift's
String.contains(), .hasPrefix(), and .hasSuffix() perform Unicode
grapheme-cluster comparison with normalization, which is both
semantically incorrect for OPA and significantly more expensive.

Changes:
- contains: haystack.contains(needle) → haystack.range(of: needle, options: .literal)
  Foundation's .literal option performs byte-level comparison without
  Unicode normalization.
- startsWith: search.hasPrefix(base) → search.utf8.starts(with: base.utf8)
- endsWith: search.hasSuffix(base) → suffix/elementsEqual on UTF-8 views

### DIFF
--- a/Sources/Rego/Builtins/Strings.swift
+++ b/Sources/Rego/Builtins/Strings.swift
@@ -59,7 +59,9 @@ extension BuiltinFuncs {
         if needle.isEmpty {
             return .boolean(true)
         }
-        return .boolean(haystack.contains(needle))
+        // Use .literal for byte-level comparison (no Unicode normalization),
+        // matching Go's strings.Contains() semantics.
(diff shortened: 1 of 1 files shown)


arirubinstein:
I think there's more juice to squeeze here with memmem - closing for now

PR #104, opened 2026-02-26
perf: use reference-type backing for Locals to eliminate CoW overhead

arirubinstein:

Replace the value-type Array backing in Locals with a reference-counted
LocalsStorage class. This eliminates copy-on-write overhead when Locals
are saved/restored across function calls in the evaluator.

Why: The IR evaluator saves and restores Locals on every function call
via the locals pool. With a value-type Array, every assignment triggers
COW reference counting checks, and any mutation after assignment causes
a full array copy. Since ownership is already managed explicitly by the
pool in IREvaluationContext, COW safety is redundant overhead.

Changes:
- Add `LocalsStorage` class as reference-counted backing store
- Locals struct now wraps `LocalsStorage` instead of `[AST.RegoValue?]`
- Pool now stores/returns `LocalsStorage` directly, avoiding struct copies
- `clearForReuse` replaces `releaseStorage` (mutates in-place, no return)
- `nonisolated(unsafe)` is safe: Locals are only accessed by a single `IREvaluationContext` at a time, never shared across threads

### DIFF
--- a/Sources/Rego/IREvaluator.swift
+++ b/Sources/Rego/IREvaluator.swift
@@ -165,8 +165,8 @@ internal final class IREvaluationContext {
     var results: ResultSet
     var locals: Locals = Locals(repeating: nil, count: 2)
 
-    // Pool for reusing storage arrays across function calls
-    private var localsPool: [[AST.RegoValue?]] = []
+    // Pool for reusing LocalsStorage objects across function calls
+    private var localsPool: [LocalsStorage] = []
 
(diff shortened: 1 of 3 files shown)


shomron:
> Can you look at the benchmarks? Comparing to baseline (main), I see up to 15% more allocations and 10% slowdown throughout the benchmarks.

I'd like to reinforce this. Performance-oriented changes like this should be backed by supporting benchmark data. @arirubinstein what prompted this change from your end?

PR #105, opened 2026-02-26
perf: cache InvocationKey hash values at construction time

arirubinstein:

Pre-compute and cache the hash value for `InvocationKey` when it is
constructed, rather than re-hashing funcName and args on every
dictionary lookup.

Why: `InvocationKey` is used as the key for the memoization cache
(`MemoCache`). Each memo lookup calls `hash(into:)`, which traverses
the `funcName` string and the entire args array of Operands. Since
`InvocationKey` is immutable (all let fields) and typically looked up
multiple times per construction, hashing the same data repeatedly is
pure overhead.

By computing the hash once in init() and returning it from
`hash(into:)`, dictionary lookups become O(1) integer operations
instead of O(n) string+array traversals.

### DIFF
--- a/Sources/Rego/IREvaluator.swift
+++ b/Sources/Rego/IREvaluator.swift
@@ -12,6 +12,20 @@ let localIdxData = Local(1)
 struct InvocationKey: Hashable {
     let funcName: String
     let args: [IR.Operand]
+    private let cachedHashValue: Int
+
+    init(funcName: String, args: [IR.Operand]) {
+        self.funcName = funcName
+        self.args = args
+        var hasher = Hasher()
+        hasher.combine(funcName)
+        hasher.combine(args)
+        self.cachedHashValue = hasher.finalize()
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(cachedHashValue)
+    }
 }
 
 /// MemoCache is a memoization cache of plan invocations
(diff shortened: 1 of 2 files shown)


koponen (dismissed):
LGTM!

Sources/Rego/IREvaluator.swift
```diff
@@ -12,6 +12,20 @@ let localIdxData = Local(1)
...
+        self.args = args
+        var hasher = Hasher()
+        hasher.combine(funcName)
+        hasher.combine(args)
+        self.cachedHashValue = hasher.finalize()
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(cachedHashValue)
+    }
```
koponen: would this benefit from overriding equality implementation and using cachedHashValue in that as a first check?

koponen (dismissed):

philipaconrad:
@arirubinstein I've assigned myself on this PR so that it won't go away on my notifications until it closes or merges. I'll see about getting a review pass in this afternoon.

philipaconrad (approved):
Thanks @arirubinstein for putting this PR together! 😄 

I greatly appreciate the test cases (makes it easier to know things work as intended), and I agree with dropping the custom `Equatable` implementation-- it's subsumed by the `Hashable` implementation that's already there.

philipaconrad:
@koponen Could you take a look at this PR, and see if we still need it?

koponen:
> @koponen Could you take a look at this PR, and see if we still need it?

It seems less critical now as the current version of this structure has merely three (u)int32s in it. Let's re-evaluate if the hashing still pops up.

philipaconrad:
Thanks @koponen, that makes sense. Thanks @arirubinstein for putting together the original PR, but I think we'll close this one for the time being, and resurrect it if hashing becomes an issue again in profiles.

PR #109, opened 2026-02-26
perf: cache the hash calculation for Operand

arirubinstein:

Operand: Pre-computes the hash value to avoid repeated hashing of OpType (String-backed RawRepresentable) and Value in hot paths like InvocationKey construction.

Every InvocationKey.init call (on every evalCall) now hashes the args array by combining pre-computed ints instead of rehashing String-backed enums.  [Operand].hash(into:) goes from O(n * string_hash) to O(n * int_combine)

### DIFF
--- a/Benchmarks/Sources/RegoBenchmarks.swift
+++ b/Benchmarks/Sources/RegoBenchmarks.swift
@@ -199,6 +199,32 @@ let benchmarks: @Sendable () -> Void = {
         }
     }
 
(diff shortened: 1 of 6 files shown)


shomron (changes requested):
Hi @arirubinstein - I have concerns with the mutability of the type leading to subtle bugs where the cached hash value no longer maps to its identity.

PR #112, opened 2026-03-03
perf: optimize string builtins with utf8 byte compare

arirubinstein:

Optimizes `strings.contains`, `strings.endswith`, and `strings.startswith` builtins to use direct UTF-8 byte comparison (`memcmp`) instead of Swift's default String methods. 

This matches Go's `strings.Contains`/`HasPrefix`/`HasSuffix` semantics - raw byte comparison with no Unicode normalization - and improves performance by avoiding Foundation overhead. 

Added a comprehensive test suite covering multi-byte UTF-8 (emoji, CJK), Unicode normalization edge cases, error handling, and boundary conditions.

### DIFF
--- a/Sources/Rego/Builtins/Strings.swift
+++ b/Sources/Rego/Builtins/Strings.swift
@@ -59,7 +59,57 @@ extension BuiltinFuncs {
         if needle.isEmpty {
             return .boolean(true)
         }
-        return .boolean(haystack.contains(needle))
+        // Use byte-level comparison (no Unicode normalization),
+        // matching Go's strings.Contains() semantics.
+        return .boolean(utf8Contains(haystack: haystack, needle: needle))
+    }
+
+    /// Byte-level substring search on raw UTF-8 buffers using `memcmp`.
+    /// Semantically identical to Go's `strings.Contains()` → `bytealg.IndexString()`.
+    private static func utf8Contains(haystack: String, needle: String) -> Bool {
+        let needleCount = needle.utf8.count
+        let haystackCount = haystack.utf8.count
+        if needleCount > haystackCount {
+            return false
+        }
+
+        // Fast path: contiguous UTF-8 storage (nearly always available for native Swift strings)
+        if let result = haystack.utf8.withContiguousStorageIfAvailable({ hBuf in
+            needle.utf8.withContiguousStorageIfAvailable({ nBuf in
+                utf8ContainsBytes(hBuf, nBuf)
+            })
+        }), let inner = result {
+            return inner
+        }
+
+        // Fallback: copy to contiguous arrays
+        // (still faster than the Foundation String.range(of:) path)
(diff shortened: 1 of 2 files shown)


Sources/Rego/Builtins/Strings.swift
```diff
@@ -97,7 +147,30 @@ extension BuiltinFuncs {
...
+            return .boolean(true)
+        }
+        let baseCount = base.utf8.count
+        let searchCount = search.utf8.count
+        if baseCount > searchCount {
+            return .boolean(false)
+        }
+
+        // Fast path: contiguous memcmp
+        if let result = search.utf8.withContiguousStorageIfAvailable({ sBuf in
```
philipaconrad, line 161: [comment]: I remember seeing this `Array<T>` [method](https://developer.apple.com/documentation/swift/array/withcontiguousstorageifavailable(_:)) when I was last browsing around the Apple Developer docs, and it's nice to see a live use case for it! 🙂

Sources/Rego/Builtins/Strings.swift
```diff
@@ -97,7 +147,30 @@ extension BuiltinFuncs {
...
+            base.utf8.withContiguousStorageIfAvailable({ bBuf in
+                guard let sBase = sBuf.baseAddress, let bBase = bBuf.baseAddress else {
+                    return false
+                }
+                return memcmp(sBase + (sBuf.count - bBuf.count), bBase, bBuf.count) == 0
```
philipaconrad, line 166: [comment]: I had to stare at this line for long time to understand why the `search` string is getting the pointer bump. The reason is because (IMO) we chose horrifically bad parameter names, and I was thinking "search" was the suffix to search for. 🤦 

Names like "search" and "suffix" might have been vastly better choices for the parameter names, but that's an upstream OPA problem, and no fault of your code. I think it should be correct as written here.

This is also the part of the code I was most worried about and thus didn't want to review hastily. Apologies about the delay on reviewing this bit!

philipaconrad (approved):
Thank you @arirubinstein for putting this PR together!

It took me a hot second to go dig up the docs for some of the parts I was not familiar with, but I have confidence now that this PR should not break anything, and does exactly what's intended. It should give us a healthy speed boost on string operations too! 😄 

Review commentary below, no changes or revisions needed:

philipaconrad:
ℹ️ Rebasing the PR through the Github Web UI (to catch up with `main`) before merging...

philipaconrad:
I'm not usually a fan of `users.noreply.github.com` emails being used in signoffs-- it's literally "click a button on github to become unreachable"-- but given that OPA has merged over 200 commits with emails of that form in the signoff, I guess the precedent has already been set. 🤷‍♂️ 


The CNCF doesn't overly much seem to care either, based on what I've been able to [dig up](https://github.com/cncf/foundation/issues/383), so I think we should be good for merge.

PR #113, opened 2026-03-03
ci(perf): Refactor benchmarks to run in CI, and add memo benchmark

arirubinstein:

This PR does the following:
* Refactors the benchmarks file to be a bit more DRY and adds two tuning knobs to help de-noise automated benchmarks at runtime.
* Adds a memoization benchmark for overlapping rules
* Introduces a github pipeline to run differential PR benchmarks which posts and updates a comment in the PR with detailed run information [example comment/pr](https://github.com/arirubinstein/swift-opa/pull/3)
* Bumps benchmark dependency to 1.31.0 to pick up jemalloc flags and linux fixes

Notes:
jemalloc in github runners are problematic, so I introduced ARC allocations as a proxy (which coincidentally the swift benchmark framework uses as a proxy when jemalloc doesn't work)

### How to test
`make perf`

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,173 @@
+name: Benchmark
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "PR number to benchmark against its base"
+        type: number
+        required: true
+      debug_benchmarks:
+        description: "Print benchmark results for debugging"
+        type: boolean
+        default: false
+
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ inputs.pr_number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  benchmark:
+    name: benchmark comparison (vs base)
+    runs-on: ubuntu-latest
+    # use swift upstream images to stay on the latest GA version
+    container: swift:noble
+    continue-on-error: true
+    env:
+      BENCHMARK_MAX_DURATION_SECONDS: "3"
+      BENCHMARK_WARMUP_ITERATIONS: "100"
+    permissions:
+      contents: read
+      pull-requests: write
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Detect Swift version
+        id: swift
+        run: echo "version=$(swift --version | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)" >> "$GITHUB_OUTPUT"
+        # jemalloc may eventually go away after swift 6.3.0 GAs due to https://github.com/ordo-one/package-benchmark/pull/333
+      - name: Install jemalloc and zstd for cache
+        run: |
+          apt-get update -q
+          apt-get install -yq libjemalloc-dev zstd jq curl
+      - name: Resolve PR refs
+        id: pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          PR_JSON=$(curl -sf \
+            -H "Authorization: token ${GH_TOKEN}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${GH_REPO}/pulls/${{ inputs.pr_number }}")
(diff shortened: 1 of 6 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,171 @@
...
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  benchmark:
+    name: benchmark comparison (vs base)
+    runs-on: ubuntu-latest
+    container: swift:noble
```
philipaconrad, line 27: [question]: Why are we running in a container for this workflow, versus just using the default Swift toolchain on the Actions runner like in other workflows? 🤔

I expect the relative diff between runs should be fine for benchmarking purposes when the performance change is large, but the precision probably is going to suffer from the virtualization overheads, right?

Benchmarks/Sources/RegoBenchmarks.swift
```diff
@@ -3,263 +3,153 @@ import Benchmark
...
-    Benchmark.defaultConfiguration.timeUnits = .nanoseconds
-
-    // Benchmark runs from the Benchmarks directory, so paths are relative to parent
-    let simpleBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/simple-directory-bundle")
-    let dynamicCallBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/dynamic-call-bundle")
-    let arrayIterationBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/array-iteration-bundle")
-    let numericLiteralsBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/numeric-literals-bundle")
-    let arrayBuildBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/array-build-bundle")
-    let objectBuildBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/object-build-bundle")
-    let setBuildBundleURL = URL(fileURLWithPath: "../Tests/RegoTests/TestData/Bundles/set-build-bundle")
```
philipaconrad, line 16: [comment]: Cleaning this up is just glorious. ✨

philipaconrad (commented):
Thanks @arirubinstein for the new workflow!

Only one or two review questions below:

philipaconrad:
@arirubinstein I've assigned myself on this PR so that it won't go away on my notifications until it closes or merges. I'll check back for another review pass later this afternoon, and at least daily thereafter until the PR lands.

.github/workflows/benchmark.yaml
arirubinstein: I chose this route since it's less ongoing maintenance (automatically latest swift, and jemalloc flows work as expected).  The macos jemalloc is mismatched to the host from brew and may potentially break down the road.

philipaconrad: Interesting! I was not aware of the mismatch issue.

Would you mind throwing a comment in somewhere above the job explaining that situation briefly? (If there's a relevant hyperlink about the problem, that'd be great to add as well for future maintainers.)

I think your maintenance concern makes sense for a manually-run job like this one that could go a while between runs. For the main CI workflows, those are exercised often enough that any breakages should become immediately and painfully visible.

arirubinstein: I can - It's due to the jemalloc being compiled for an earlier macOS than the one that it's being installed in due to the brew source, and possibly related to the jemalloc maintenance stuff which should hopefully be resolved (https://engineering.fb.com/2026/03/02/data-infrastructure/investing-in-infrastructure-metas-renewed-commitment-to-jemalloc/)
At the same time, swift benchmark is moving away from jemalloc, so this may become moot as well

philipaconrad (dismissed):
Thanks @arirubinstein for this new GH Actions workflow!

I'd appreciate a small comment somewhere about the `jemalloc` situation, but it's not a blocker for merging this in my view.

I'm looking forward to seeing the benchmarking job in action! 😄

arirubinstein:
no problem! Added two comments, including one about the jemalloc potentially going away

philipaconrad (approved):

philipaconrad:
ℹ️ Rebasing via the Github Web UI to catch up with `main`. Will merge when green...

Issue #116, opened 2026-03-14
builtins: implement remaining time builtins


DFrenkel:
### What is the underlying problem you're trying to solve?
Now that `time.now_ns` has been implemented such that is returns the same value within a single policy evaluation query, we need implementation of the remaining [time](https://www.openpolicyagent.org/docs/policy-reference/builtins/time) builtins:
* `time.add_date`
* `time.clock`
* `time.date`
* `time.diff`
* `time.format`
* `time.parse_duration_ns`
* `time.parse_ns`
* `time.parse_rfc3339_ns`
* `time.weekday`

### Expected Outcome
* `time` builtins can be used in policy evaluations
* All *time* tests pass: https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/time

PR #120, opened 2026-03-16
builtins: add time.clock, time.date, and time.weekday

DFrenkel:
The following builtins have been implemented:
* [time.clock](https://www.openpolicyagent.org/docs/policy-reference/builtins/time#builtin-time-timeclock)
* [time.date](https://www.openpolicyagent.org/docs/policy-reference/builtins/time#builtin-time-timedate)
* [time.weekday](https://www.openpolicyagent.org/docs/policy-reference/builtins/time#builtin-time-timeweekday)

### Definition of done
Compliance tests pass:
```
✅ time/test-time-0951.json: time/date -> eval
✅ time/test-time-0952.json: time/date with LA tz -> eval
✅ time/test-time-0953.json: time/date with empty tz -> eval
✅ time/test-time-0954.json: time/date leap day -> eval
✅ time/test-time-0955.json: time/date too big -> eval
✅ time/test-time-0956.json: time/clock -> eval
✅ time/test-time-0957.json: time/clock with NY tz -> eval
✅ time/test-time-0958.json: time/clock leap day -> eval
✅ time/test-time-0959.json: time/clock too big -> eval
✅ time/test-time-0960.json: time/weekday -> eval
✅ time/test-time-0961.json: time/weekday -> eval
✅ time/test-time-0962.json: time/weekday -> eval
✅ time/test-time-0963.json: time/weekday -> eval
✅ time/test-time-0964.json: time/weekday -> eval
✅ time/test-time-0965.json: time/weekday -> eval
✅ time/test-time-0966.json: time/weekday -> eval
✅ time/test-time-0967.json: time/weekday too big -> eval
```

### How to test
```
make all
OPA_COMPLIANCE_TESTS="time" make test-compliance | grep -E "weekday|clock|date"
```

### Related Resources
Related to #116

### DIFF
--- a/Sources/Rego/Builtins/Time.swift
+++ b/Sources/Rego/Builtins/Time.swift
@@ -116,7 +116,11 @@ extension BuiltinFuncs {
     /// a UTC-based Gregorian `Calendar` configured with the specified timezone.
     private static func dateWithCalendar(_ arg: AST.RegoValue) throws -> (Date, Calendar) {
         let (ns, timeZone) = try parseNanosWithTimezone(arg)
-        let date = Date(timeIntervalSince1970: TimeInterval(ns / 1_000_000_000))
+        // Use floor division rather than truncation.
+        // Without this, negative timestamps with sub-second nanos land in the wrong second,
+        // e.g. ns=-900_000_000 (-0.9s) would truncate to 0 (epoch) instead of -1 (1969-12-31 23:59:59).
+        let wholeSeconds = ns / 1_000_000_000 - (ns % 1_000_000_000 < 0 ? 1 : 0)
+        let date = Date(timeIntervalSince1970: TimeInterval(wholeSeconds))
         var cal = Calendar(identifier: .gregorian)
         cal.timeZone = timeZone
         return (date, cal)
--- a/Sources/Rego/Builtins/Time.swift
+++ b/Sources/Rego/Builtins/Time.swift
@@ -3,8 +3,10 @@ import Foundation
 
 private let minDateAllowedForNsConversion = Date(timeIntervalSince1970: TimeInterval(Int64.min) / 1_000_000_000)
 private let maxDateAllowedForNsConversion = Date(timeIntervalSince1970: TimeInterval(Int64.max) / 1_000_000_000)
+private let weekdayNames = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
 
 extension BuiltinFuncs {
+    // MARK: - time.now_ns
     static func timeNowNanos(ctx: BuiltinContext, args: [AST.RegoValue]) async throws -> AST.RegoValue {
         guard args.count == 0 else {
             throw BuiltinError.argumentCountMismatch(got: args.count, want: 0)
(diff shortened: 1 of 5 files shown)


Sources/Rego/Builtins/Time.swift
```diff
@@ -71,6 +74,96 @@ extension BuiltinFuncs {
...
+        let (date, cal) = try dateWithCalendar(args[0])
+        return .string(weekdayNames[cal.component(.weekday, from: date) - 1])
+    }
+
+    // MARK: - private utilities
+    /// Parses `nanoseconds` or `[nanoseconds, timezone]` argument and returns the corresponding `Date` and
+    /// a UTC-based Gregorian `Calendar` configured with the specified timezone.
+    private static func dateWithCalendar(_ arg: AST.RegoValue) throws -> (Date, Calendar) {
+        let (ns, timeZone) = try parseNanosWithTimezone(arg)
+        let date = Date(timeIntervalSince1970: TimeInterval(ns / 1_000_000_000))
```
philipaconrad: [comment]: I'd wondered why the ns timestamp was truncated so severely here. It turns out [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval) expects a timestamp in *seconds*, so the truncation makes sense in context. For date/calendar operations, I *think* that should be fine for our purposes.

I'd be curious to see an edge-case test for this though-- I'm not sure what the exactly correct behavior should be, for us or OPA. 🤔 If a timestamp is technically 9/10 of a second into the new year, we'd want to handle that case, right?

Tests/RegoTests/BuiltinTests/TimeTests.swift
```diff
@@ -98,9 +406,24 @@ extension BuiltinTests.TimeTests {
...
+                argName: "x",
+                allowedArgTypes: ["array", "number[integer]"],
+                wantArgs: "any<number[integer], array<number[integer], string>>",
+                generateNumberOfArgsTest: true, numberAsInteger: true),
+            BuiltinTests.generateFailureTests(
+                builtinName: "time.date", sampleArgs: [[1_585_852_421_593_912_000, "UTC"]], argIndex: 0,
+                argName: "x",
+                allowedArgTypes: ["array", "number[integer]"],
+                wantArgs: "any<number[integer], array<number[integer], string>>",
+                generateNumberOfArgsTest: true, numberAsInteger: true),
```
philipaconrad, line 462: [question]: Could we add the `generateFailureTests` for `time.weekday` as well?

philipaconrad (commented):
Hello @DFrenkel! Overall, this PR looks great!

I've got a few comments and questions below:

Sources/Rego/Builtins/Time.swift
DFrenkel: If we get sub-second value of the timestamp, truncating it to a whole second should *not* change either the [year, month, date] or [hour, minute, second] value of it, and the weekday value, should it? I added tests for sure, thanks for pointing this out!
Also: turns out there *was* a bug in this for handling negative timestamps (before epoch), so this is also updated

Tests/RegoTests/BuiltinTests/TimeTests.swift
DFrenkel: ✅ Good catch - this was my miss.

philipaconrad:
@DFrenkel I've assigned myself on this PR so that it won't go away on my notifications until it closes or merges.

Sources/Rego/Builtins/Time.swift
philipaconrad: I think the truncation behavior sounds reasonable, especially in context of `time.clock` / `time.date` / `time.weekday`, which are somewhat coarse/lossy ways of looking at a timestamp anyhow. Thanks for taking a second look at the timestamp handling!

philipaconrad (approved):
Thanks @DFrenkel for taking a second look at the timestamp handling, and for adding some extra edge-case tests. I think we're good to merge! 🙂

PR #124, opened 2026-03-17
IR+StaticAnalysis: Add locals renumbering pass

### LINKED ISSUE #115: IR Static Analysis: Register (Local) Renumbering Pass
## Motivation / Problem

Currently, if a malicious upstream policy source wants to use excessive memory (or cause an OOM!) in a Swift OPA IR evaluator, all they have to do is simply add an IR instruction into a Plan with a ridiculous `Local` index, like so:

```
    AssignIntStmt  int64(0),  2147483647
//  ^^^^^^^^^^^^^  ^^^^^^^^   ^^^^^^^^^^
//  instr          source     target
```

In any IR plan, we _should_ be able to trivially rewrite the plan so that all of the `Local` indices (register numbers) are contiguous within a given IR Plan/Func. This can be done in 1-2x linear scans over the IR instructions in the plan.

# Proposed Solution

We should add a new IR static analysis pass that runs *before* (or as part of!) the `computeMaxLocal` function. The new pass will do the 1-2x scans over the IR, tracking seen indexes, and performing index remapping as necessary.

The end result will have a contiguous set of `Local` indexes across the Plan/Func.

This strongly bounds how many "dead indexes" are possible when we allocate arrays later on for Locals storage in the IR evaluator.

## Single-scan, naive remap
A naive method would be to use a counter variable and a dictionary of `[Int: Int]`, bumping up the counter variable as new contiguous indexes are seen in the policy. As remappings occur, the dictionary accrues more entries. This could result in excessive remapping in some cases.

## 2-scan, cleaner remap
A less naive, 2-pass method would be to scan over the instructions in the Plan/Func, and track every index seen, then a second pass would perform the remapping optimally, avoiding spurious remappings at lower index values.

## Related Resources

 - [Register renaming](https://en.wikipedia.org/wiki/Register_renaming): A range of hardware techniques for dynamically remapping registers at execution time.
 - [Register allocation](https://en.wikipedia.org/wiki/Register_allocation): A field of compiler techniques for mapping N-many logical registers onto a limited set of physical registers.

# How to test?

Tests could be built around fuzzers generating/mutating the indexes in normal policies, and would check that the final list of "seen indexes" in the renumbered IR are contiguous.

philipaconrad:

This PR adds a renumbering pass to our IR static analysis that ensures all locals in a policy's IR statements are of a contiguous range.

This prevents cases where a local is set to a random large value (e.g. 1234567) from hurting our locals preallocation in the evaluator. After renumbering a plan or func, the max local index in its IR statements will be equal to the number of unique locals present in that plan or func.

We can prealloc the locals array in peace now! 😅 

~~Note: The current test case is ugly as heck, and large, due to the size of the `Policy` type initializers. I'll see about cleaning it up a bit, possibly by converting the `Policy` arguments to JSON, and just adding some comments about what's different between the input and expected result.~~

Note 2: It may be worth merging the renumbering and `maxLocal` logic together. @koponen, let me know if you think it's worthwhile.

Fixes: #115

---

Update: I wound up reworking this PR to the 2-pass algorithm, because the encounter-ordering of the 1-pass algorithm resulted in some very arbitrary remapping, and resulted in needless churn, even on "compact" (already contiguous locals) IR policies.

### Definition of done

 - Renumbering works.
 - "Easy" test cases work.
 - [x] Add an edge case test or two?

### How to test

 - There's a new test suite under `Tests/IRTests/IRStaticAnalysisTests.swift`. This is picked up automatically by `make test`.

### DIFF
--- a/Sources/Rego/VM.swift
+++ b/Sources/Rego/VM.swift
@@ -169,7 +169,9 @@ extension VM {
                 return BlockResult(breakCounter: UInt32(length))
             case Int(Opcode.assignVar1.rawValue):
                 // Payload is [target:12 | source:12] packed in the header length field.
-                context.locals[Local(UInt32(length) >> 12)] = context.locals[Local(UInt32(length) & 0xFFF)]
+                // Read through resolveLocal so an unbound source is promoted to .undefined,
+                // matching the non-compact assignVar path.
+                context.locals[Local(UInt32(length) >> 12)] = context.resolveLocal(idx: Local(UInt32(length) & 0xFFF))
                 continue
             // ── Non-compact opcodes (rawValues 0–34): `length` is payload byte count ─────────
             // Assignments
--- a/Sources/IR/IR+Decode.swift
+++ b/Sources/IR/IR+Decode.swift
@@ -15,15 +15,31 @@ extension Policy {
     /// Prepare the policy for execution by running static analysis passes.
     /// This computes properties like maxLocal that are used for optimization.
     public mutating func prepareForExecution() throws {
-        // Compute maxLocal for all plans
+        // Renumber locals in every plan and func to compact contiguous ranges.
+        if var plans = self.plans {
+            for i in plans.plans.indices {
+                plans.plans[i].renumberLocals()
+            }
+            self.plans = plans
+        }
+        if var funcs = self.funcs {
+            if var funcList = funcs.funcs {
+                for i in funcList.indices {
+                    funcList[i].renumberLocals()
+                }
+                funcs.funcs = funcList
+            }
+            self.funcs = funcs
+        }
+
+        // Compute maxLocal after renumbering so the cached value reflects the
+        // compacted range used by the evaluator.
         if var plans = self.plans {
             for i in plans.plans.indices {
                 plans.plans[i].computeMaxLocal()
             }
             self.plans = plans
         }
-
-        // Compute maxLocal for all functions
         if var funcs = self.funcs {
             if var funcList = funcs.funcs {
                 for i in funcList.indices {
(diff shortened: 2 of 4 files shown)


koponen:
The static analysis passes are fairly cheap, not sure their merging is worth the effort unless their complexity really remains the same.

koponen:
Is this more defensive or have you seen the planner leave gaps in valid plans?

philipaconrad:
> Is this more defensive or have you seen the planner leave gaps in valid plans?

This is entirely defensive at present. I haven't seen the planner leave "gaps" in plans it emits.

But, I can imagine that in the future we may have more complex optimization passes, or even non-OPA IR sources upstream (imagine if FGA or similar RBAC systems emitted IR plans, for example) that might result in accidental or intentional "gaps" in locals numbering. This pass would be a relatively cheap way to ensure that even a malicious choice of locals numbering can't cause issues.

Sources/Rego/VM.swift
```diff
@@ -169,7 +169,9 @@ extension VM {
                 return BlockResult(breakCounter: UInt32(length))
             case Int(Opcode.assignVar1.rawValue):
                 // Payload is [target:12 | source:12] packed in the header length field.
-                context.locals[Local(UInt32(length) >> 12)] = context.locals[Local(UInt32(length) & 0xFFF)]
+                // Read through resolveLocal so an unbound source is promoted to .undefined,
+                // matching the non-compact assignVar path.
+                context.locals[Local(UInt32(length) >> 12)] = context.resolveLocal(idx: Local(UInt32(length) & 0xFFF))
```
philipaconrad, line 174: @koponen This is the bit I'd like you to look at-- if the new form with `context.resolveLocal` was the intention from the beginning, I can drop the explainer comment.

koponen: Yes, let's keep the two implementations in sync - it's a bug they are not the same. I'm actually preparing a patch that will optimize the undefined/nil separation more broadly so we can return to this then.

koponen (approved):
LGTM

PR #125, opened 2026-03-19
Rego/Manifest: Add `Codable` implementation.

philipaconrad:

This PR extends the `OPA.Manifest` type with a `Codable` implementation. This allows easily extending support for different structured data formats in the future (e.g. YAML).

As an example of how useful this is, our previously custom JSON decoding logic becomes a one-liner in the PR, and YAML support in the Swift OPA SDK will be similarly straightforward to add once this PR lands.

### How to test

 - Run existing Bundle test suite as part of `make test`. The new decoder-based logic works just as well as the original custom logic did.


srenatus (approved):
Nice! So the Codable let us use standard machinery, so we don't have to decode into a dictionary first. 👍

philipaconrad:
ℹ️ Rebasing through the Github Web UI to catch up with `main` before merging...

PR #129, opened 2026-04-16
Tracer: fix 'occured' -> 'occurred' in doc comments

SAY-5:
Two doc comments in `Sources/Rego/Tracer.swift` (lines 19, 29) read `that occured`. Fixed to `occurred`. Comment-only change.

### DIFF
--- a/Sources/Rego/Tracer.swift
+++ b/Sources/Rego/Tracer.swift
@@ -25,7 +25,7 @@ extension OPA.Trace {
         case note
     }
 
-    /// Describes the source location at which a trace event occured
+    /// Describes the source location at which a trace event occurred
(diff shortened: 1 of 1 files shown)


philipaconrad:
Thanks for catching that typo, but I only see a correction on one of the two comments?

SAY-5:
Good catch — fixed the second occurrence on the `Location.Source` comment (line 29) in the same force-push. Both doc comments now read `occurred`.

philipaconrad (approved):
Thanks for the cleanup!

philipaconrad:
ℹ️ Rebasing this branch to catch up with `main` via the Github Web UI, and then I'll merge it...

PR #133, opened 2026-04-22
perf: fix benchmark bundle path and surface prepareForEvaluation errors

koponen:


philipaconrad (approved):
This probably got busted when I merged the improved bundle overlap detection algorithm in #110. I added the extra level of nesting because I had assumed the path was computed *relative to the deeper-nested source file*, and it seems the path is relative to *where the binary is running*. 🤦 

Thanks for catching this, @koponen!

PR #135, opened 2026-04-27
feat(builtins): any_prefix_match, any_suffix_match, replace_n, render_template

### LINKED ISSUE #134: builtins: Implement remaining strings builtins: any_prefix_match, any_suffix_match, replace_n, render_template

Implement the remaining [`strings`](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings) builtins from the OPA specification:  
- [x] [strings.any_prefix_match](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings#builtin-strings-stringsany_prefix_match)                             
- [x] [strings.any_suffix_match](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings#builtin-strings-stringsany_suffix_match)
- [x] [strings.replace_n](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings#builtin-strings-stringsreplace_n)                                   
- [x] [strings.render_template](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings#builtin-strings-stringsrender_template)

### Expected Outcome

- `strings` builtins can be used in policy evaluations
- All *strings* tests pass: https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/strings

DFrenkel:
- Implement strings.any_prefix_match, strings.any_suffix_match, strings.replace_n, and strings.render_template builtins
- Fix strings.count empty substring behavior to match Go's strings.Count("", "") == 1 semantics
- Add prefix-compressed radix trie (ByteTrie) and Go text/template subset engine (TextTemplate) as standalone modules with dedicated tests

### Definition of done
Prefix/suffix matching uses a radix trie with prefix compression, modeled after OPA's patricia trie (go-patricia). Nodes store compressed byte sequence prefixes rather than single bytes, reducing node count for path-heavy policies. Insert is O(k) per key byte length k, query is O(p) per prefix byte length p — independent of the number of keys in the trie. Single-element fast path bypasses the trie entirely with memcmp-based comparison, consistent with the byte-level optimizations in contains/endsWith from prior commits.

replace_n matches the behavior of Go's strings.NewReplacer: patterns sorted lexicographically, left-to-right scan, first (lex-earliest) match wins at each position.

render_template implements a subset of Go's text/template (field access, range, if/else) with <no value> → <undefined> post-processing matching the OPA Go implementation.

### How to test
`make test` and relevant `OPA_COMPLIANCE_TESTS="strings" make test-compliance` tests pass.
Note that there are still failures related to lack of `regex.find_all_string_submatch_n` builtins, which will be fixed later

### Related Resources
Resolves #134


philipaconrad (approved):
Thanks @DFrenkel for this PR! Overall, it looks great. 😄 

I've looked over the builtin implementations and skimmed over the tests, and they all look solid. The `text/template` implementation is the bit that I'm least confident about, but it looks reasonable, and has a good battery of test cases.

I like the use of a trie for the prefix/suffix matching, and I appreciate the comments there about how the implementation differs from what Golang does-- that could be a helpful breadcrumb to follow in the future if profiling indicates we need to optimize those builtins further.

PR #136, opened 2026-04-27
rego: Add synchronous evaluation path via codegen.

koponen:
Sync-safety is determined statically at load time and sync-safety bits are injected into bytecode instructions. Plans that call async-only builtins (directly, transitively, or through callDynamic when any function in the policy calls one) throw requiresAsyncEvaluation on the sync path. Plans using callDynamic with only sync builtins are fully sync-safe.

The synchronous VM overloads are generated by tools/SyncGen, a standalone Swift tool that reads // @sync comment annotations in the source.

The optimization reduces CPU instructions by 6 - 15%, wall clock by 9 - 23%, and heap allocations by up to 38% across all benchmarks.


philipaconrad:
ℹ️ @koponen, I fixed the merge conflict caused by merging #134 a few minutes ago, and pushed the resolved commit up to the PR branch. No changes outside of `Sources/Rego/Builtins/Strings.swift`.

Sources/Bytecode/Opcode.swift
```diff
@@ -273,7 +273,7 @@ public struct InstructionHeader: Sendable {
             for _ in 0..<numBlocks {
                 guard off + 8 <= start + length else { throw Error.invalidPayloadLength }
                 let blockOffset = Int(readUInt32(payload, at: off))
-                let blockSize = Int(readUInt32(payload, at: off + 4))
+                let blockSize = Int(readUInt32(payload, at: off + 4) & 0x7FFF_FFFF)  // mask sync-safety bit (bit 31)
```
philipaconrad: [question]: Do we have this new sync-safety bit documented anywhere outside of this spot? 😅

Sources/Rego/Builtins/Registry.swift
```diff
@@ -1,9 +1,19 @@
...
+public typealias SyncBuiltin = @Sendable (BuiltinContext, [AST.RegoValue]) throws -> AST.RegoValue
+
+/// Whether a builtin is synchronous (usable on both the async and sync VM paths)
+/// or async-only (usable on the async path only, e.g. because it performs network I/O).
+public enum BuiltinImpl: Sendable {
+    case sync(SyncBuiltin)
+    case asyncOnly(AsyncBuiltin)
+}
+
+public struct BuiltinContext: @unchecked Sendable {
```
philipaconrad, line 20: [question]: Why `@unchecked Sendable` here? Is one of the members' types not explicitly `Sendable`?

Sources/Rego/Sync.swift
philipaconrad, line 1: [comment]: This is a pretty clever way of generating the right code for each path. I'm not particularly familiar with Swift's macro system yet, so this is a technique that would not have immediately occurred to me as a thing to try out. 🤔

Sources/SyncMacroPlugin/SyncMacro.swift
philipaconrad, line 1: [comment]: This is effectively magic to me. I'm not familiar enough with Swift macros to be able to critique this macro effectively, so I'll trust your judgement here. 😅

philipaconrad (approved):
Thank you for the interesting PR, @koponen! The changes I see across the VM / Evaluator / Engine all make sense, given how the new `@sync` macro is supposed to work. Unfortunately, I can't really critique the macro's implementation much, but it seems reasonable enough.

My main concern point that came up when reviewing was the new "sync-safe" bit flag that's being set in a few places down in the bytecode, but thankfully it seems that there is appropriate masking for that bit in the evaluator. My fear was that if the PR missed a bit-setting/bit-masking spot, it'd cause some potentially weird bugs down the line.

If you're feeling charitable to future maintainers, you could add comments to some of the higher-level parts (the `executeX` funcs that call down into `execX` funcs), specifying that the bit masking happens further down in the call chain. That took me a sec to verify when I was perusing the sources. 😅 

I have just a few questions and comments below, but overall, this looks good! 😄

ComplianceSuite/Sources/RegoCompliance/RegoCompliance.swift
```diff
@@ -202,7 +202,114 @@ public struct ComplianceTesting {
         public var knownIssue: String?
     }
 
-    public static func runTest(config: TestConfig, _ tc: IRTestCase, _ customBuiltins: [String: Builtin] = [:]) async
+    /// Returns true when the test case references `test.sleep`, which is an async-only builtin
+    /// with no sync counterpart and therefore cannot run through the synchronous evaluation path.
+    public static func requiresAsyncBuiltins(_ tc: IRTestCase) -> Bool {
+        return (tc.base.modules ?? []).contains { $0.contains("test.sleep") }
```
philipaconrad, line 208: This seems a little brittle, but it's fine for purposes of this specific compliance test suite.

Sources/Rego/Builtins/Registry.swift
```diff
@@ -53,16 +67,54 @@ public struct BuiltinContext {
...
     public static var defaultRegistry: BuiltinRegistry {
-        BuiltinRegistry(
-            builtins: BuiltinRegistry.defaultBuiltins
-        )
+        BuiltinRegistry(builtins: defaultBuiltins.mapValues { .sync($0) })
     }
 
-    internal static var defaultBuiltins: [String: Builtin] {
+    /// Creates a registry by merging custom async-only builtins with the default sync builtins.
+    /// Custom builtins are available on the async path only.
```
philipaconrad: [question]: Why are custom builtins forced to be async? Would it be painful to support custom sync builtins? 🤔

Sources/Rego/Engine.swift
```diff
@@ -21,14 +21,18 @@ extension OPA {
...
 
         // Bundle-owned `/data` subtrees written by the most recent
         // `prepareForEvaluation(query:)` call. The next prepare call uses
         // this set (along with the new prepare's roots) to decide which
         // subtrees to erase before writing fresh bundle data into the store.
         // User data outside this set survives across prepare calls.
         private var lastWrittenBundleRoots: Set<StoreKeyPath> = []
+
+        /// When `false`, skip `SyncSafePatcher` at load time so the async VM path is
+        /// always taken during evaluation.  Intended for testing only.
```
philipaconrad:
```suggestion
        /// always taken during evaluation. Intended for testing only.
```
[nit]

tools/SyncGen/Sources/SyncGen/main.swift
```diff
@@ -0,0 +1,99 @@
...
+// Build output.
+var output = """
+    // Generated by tools/SyncGen — do not edit manually.
+    // To regenerate:
+    //   swift run --package-path tools/SyncGen SyncGen \\
+    //       Sources/Rego/VM.swift Sources/Rego/VM+Instructions.swift \\
+    //       Sources/Rego/Engine.swift Sources/Rego/IREvaluator.swift \\
+    //       --output Sources/Rego/Generated/SyncPeers.swift
+
+    import AST
```
philipaconrad, line 71: [comment]: I do feel like the fixed imports here may be a bit brittle, but I can't think of an easier way to get the same result at the moment. This should be fine for our purposes-- if we need something more robust later, we'll know immediately, thanks to compile errors.

tools/SyncGen/Sources/SyncGen/Rewriter.swift
```diff
@@ -0,0 +1,84 @@
...
+        pieces.append(piece)
+    }
+    return Trivia(pieces: pieces)
+}
+
+// Returns the if-branch body when `item` is a sync-dispatch if/else whose condition
+// matches one of the known sync-safety checks. Handles both .expr(IfExprSyntax) and
+// .stmt(ExpressionStmtSyntax{IfExprSyntax}) — the latter is how swift-syntax 600.0.x
+// represents if/else statements in function bodies.
+func syncDispatchIfBody(_ item: CodeBlockItemSyntax) -> CodeBlockItemListSyntax? {
```
philipaconrad, line 63: [comment]: I'd wondered how we were pruning the async branches in the VM. This is a little gross, but I can't argue with how effective it is. Again, if we need something more robust down the road, we'll know pretty quickly. 😅

philipaconrad (commented):
Overall, this looks good! 😄 

Thanks for carving off the macro into a separate package/module so that downstream users won't need to bother with `swift-syntax`.

I have a few review comments below. The most critical is about why we're limiting custom builtins to be async-only, as the machinery looks reasonable for allowing *either* sync or async custom builtins to be added.

Once we've wrapped up those comments, I'm satisfied that this changeset should be good to merge. 🙂

Sources/Rego/Builtins/Registry.swift
koponen: That's a mere simplification. I'll add support for custom sync builtins.

philipaconrad (approved):
Thank you @koponen for adding the extra wiring to make `sync` custom builtins possible! This looks great, and I think we're good to merge!

PR #146, opened 2026-05-11
Revert "Engine: Add MiniPlanner for basic data queries."

philipaconrad:
## What code changed, and why?

This PR reverts open-policy-agent/swift-opa#145, for the following reasons:

 - The inconsistencies with how querying all of `data` works vs OPA can cause problems/bugs/confusion for users, **even if we document the behavior prominently**.
   - Related question: Do we want to standardize on this behavior across IR interpreters (e.g. [`open-policy-agent/java-opa-sdk`](https://github.com/open-policy-agent/java-opa-sdk))? Talking with other maintainers (@tsandall, @srenatus, @johanfylling) indicates probably not for now.
 - @johanfylling raised the point that it basically turns the `OPA.Engine` into an oracle for everything in `data`, which was kind of the point, but is definitely unexpected behavior in an IR policy context, where things are normally scoped to _just_ a limited set of planned entrypoints.

Based on the above points, as well as the fact that the `MiniPlanner` was built out specifically to enable [Discovery](https://www.openpolicyagent.org/docs/management-discovery) bundles (often JSON-only, no policy) to work, it makes sense to just lift this feature out of Swift OPA, and move it over into the Swift OPA SDK, where it can be scoped down to supporting the discovery decision path.

This migration will have the benefit of preventing unexpected new behaviors for Swift OPA users, while still unblocking [Discovery](https://www.openpolicyagent.org/docs/management-discovery) in the SDK.


philipaconrad:
ℹ️ Apparently I'm able to force-push over a Github-generated revert branch because I'm a repo admin. I've updated only the commit message. The code changes are the same as before.

srenatus (approved):

PR #147, opened 2026-05-11
IR: Fix Encodable implementation for Block and Operand.

philipaconrad:

This commit fills in the `Encodable` implementation for the `IR.Block` type, and for `IR.Operand`. It is now possible to round-trip the IR policy JSON through the `IR` module types, and back again to JSON.

Before these changes, Blocks only printed a debug string showing the number of statements in the block, instead of the serialized IR statements. This broke JSON encoding.

The Operand type had issues from being an enum type. It now has a custom encoding implementation, ensuring the three operand types render correctly in JSON.

### How to test

 - New tests are automatically picked up by `make test`, as part of the IR "Golden" and encoding test suites.

### DIFF
--- a/Tests/IRTests/GoldenIRTests.swift
+++ b/Tests/IRTests/GoldenIRTests.swift
@@ -24,3 +24,17 @@ func testParsingGolden(input: TestURL) async throws {
     let data = try Data(contentsOf: input.url)
     let _ = try JSONDecoder().decode(Policy.self, from: data)
 }
+
+// Round-trip each golden fixture through the full serdes flow:
+// JSON -> Policy -> JSON -> Policy, asserting that the re-decoded
+// policy matches the first one. The JSON bytes themselves are not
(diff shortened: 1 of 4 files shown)


Tests/IRTests/GoldenIRTests.swift
```diff
@@ -24,3 +24,17 @@ func testParsingGolden(input: TestURL) async throws {
     let data = try Data(contentsOf: input.url)
     let _ = try JSONDecoder().decode(Policy.self, from: data)
 }
+
+// Round-trip each golden fixture through the full serdes flow:
+// JSON -> Policy -> JSON -> Policy, asserting that the re-decoded
+// policy matches the first one. The JSON bytes themselves are not
+// required to match (key order, whitespace, and optional/default
+// fields can differ), but the model must survive the trip.
```
srenatus, line 32: 👏

srenatus (approved):
LGTM thanks!

PR #151, opened 2026-05-29
Engine: Improve plan validation and store behavior. Add remove op for store.

philipaconrad:

This commit includes improvements to plan name validation, and adds a remove operation for the store.

The Engine now validates that plans are contained under the roots of their parent bundles and also supports loading user-provided IR policies alongside bundles now. The validation ensures that plan names don't conflict across bundles or user-provided IR policies, and prevent user IR policies from claiming paths owned by bundles.

The store also now supports user-supplied data living alongside bundle-owned data. `prepareForEvaluation` now does not erase the entire store, and only clears out the old/stale bundle roots before writing the new set of bundle data/roots to the store. Any paths outside of that set are left untouched.

The `Store` protocol now supports a remove operation as well, bringing us closer to the OPA store's functionality.

### How to test

 - New tests have been added under `Tests/ASTTests/` and `Tests/RegoTests/`, and are automatically picked up by `make test`.

### Notes for reviewers

 - Only 360 LOC (~1/3 of the PR size) is actual library code changes. The other 2/3 is new tests.

### DIFF
--- a/Sources/AST/RegoValue+PatchMerge.swift
+++ b/Sources/AST/RegoValue+PatchMerge.swift
@@ -38,6 +38,77 @@ extension RegoValue {
             return .object(o)
         }
     }
+
+    // removing returns a new RegoValue with the value at the given path removed.
+    // If the path does not exist, the value is returned unchanged. For an object
+    // node the leaf key is removed from its parent. For an array node the element
+    // at the integer-valued segment is removed and the array is compacted.
+    // Ancestors are preserved even if they become empty as a result. A segment
+    // that does not address the current node (a non-integer index into an array,
+    // an out-of-bounds index, a missing object key, or any segment into a scalar
+    // or set) is a no-op. An empty path resets the value to an empty object,
+    // matching the "reset the store" semantics callers expect at the root.
+    package func removing(at path: [String]) -> RegoValue {
+        // An empty path addresses the root itself. Removing the root resets the
+        // value to an empty object, matching the "reset the store" semantics
+        // callers expect. This only makes sense at the root, so it is handled
+        // here at the entry point rather than in the recursive core, which only
+        // ever sees non-empty paths.
+        guard !path.isEmpty else {
+            return .object([:])
+        }
+        return removing(at: path[...])
+    }
+
+    // This variant of ``removing`` works on an ArraySlice, and is the
+    // recursive core for removal operations. It assumes a non-empty path: the
+    // recursion removes a leaf directly once `restOfPath` is empty and only
+    // recurses on a non-empty remainder, so the empty-path case never reaches
+    // here. The root-reset case is handled by the entry point above.
+    package func removing(at path: ArraySlice<String>) -> RegoValue {
+        guard let i = path.indices.first else {
+            // Defensive no-op: a non-root empty path does not address anything.
+            return self
+        }
+        let segment = path[i]
+        let restOfPath = path[i.advanced(by: 1)...]
+
+        switch self {
+        case .object(var o):
+            let k = RegoValue.string(segment)
+            guard let child = o[k] else {
+                return self
+            }
+            if restOfPath.isEmpty {
+                // Leaf: remove the key from the parent.
+                o.removeValue(forKey: k)
+                return .object(o)
+            }
+            // Recurse into the child; reattach the child structure, post-deletion.
+            o[k] = child.removing(at: restOfPath)
+            return .object(o)
+
+        case .array(var a):
+            // Arrays are addressed by a non-negative integer index. Anything
+            // else (non-integer or out-of-bounds) does not address an element,
+            // so it is a no-op.
+            guard let idx = Int(segment), idx >= 0, idx < a.count else {
+                return self
+            }
+            if restOfPath.isEmpty {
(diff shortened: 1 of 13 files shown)


Sources/AST/RegoValue+PatchMerge.swift
```diff
@@ -38,6 +38,45 @@ extension RegoValue {
...
+    // become empty as a result. An empty path resets the value to an empty object,
+    // matching the "reset the store" semantics callers expect at the root.
+    package func removing(at path: [String]) -> RegoValue {
+        return removing(at: path[...])
+    }
+
+    package func removing(at path: ArraySlice<String>) -> RegoValue {
+        // Base case: empty path => reset to an empty object.
+        if path.isEmpty {
+            return .object([:])
```
koponen: I found this check here in the recursive version quite confusing, especially since you have then the check rest.isEmpty below to avoid this code path being hit during the recursion. Would it better have this logic at the entry point (the function above)?

Sources/Rego/Engine.swift
```diff
@@ -194,6 +212,33 @@ extension OPA.Engine {
...
+                newRoots.insert(StoreKeyPath(["data"] + segments))
+            }
+        }
+
+        // Erase old + new bundle-owned subtrees in lexicographic order so an
+        // ancestor is removed before any of its descendants.
+        let toErase = lastWrittenBundleRoots.union(newRoots)
+        for path in toErase.sorted(by: { $0.segments.lexicographicallyPrecedes($1.segments) }) {
+            try await store.remove(at: path)
+        }
```
koponen, line 240: What's the contract with the storage and parallelism: this store cannot be shared across multiple query preparations because of the non-atomicity of these operations.  What query is prepared for activation in the background while evaluations execute in parallel?

Tests/ASTTests/RemovingTests.swift
```diff
@@ -0,0 +1,123 @@
...
+                data: .object([
+                    .string("foo"): .number(1),
+                    .string("bar"): .number(2),
+                ]),
+                path: ["foo"],
+                expected: .object([
+                    .string("bar"): .number(2)
+                ])
+            ),
+            TestCase(
```
koponen, line 28: for completeness, should test also the removal paths going through arrays are considered no-ops.

koponen (commented):
The concurrency contract between stores and query preparing due to new bundle is the main question.

Sources/Rego/Engine.swift
philipaconrad: I'm pretty sure the Engine at present is not `Sendable`, mostly from the store layer not being `Sendable`.

I don't quite understand the second question-- I'd hope that a `PreparedQuery` can be run at any time after prep?

Sources/AST/RegoValue+PatchMerge.swift
philipaconrad: Probably a holdover from earlier stuff. I'll take a look and clean it up to do the check in one place instead of two.

Sources/Rego/Engine.swift
koponen: Here's roughly what I was going after: PreparedQuery is Sendable but the Store within is not, so a (custom) store implementation could be a reference, and therefore shared. Now multi step store operations could cause inconsistencies. Now that I'm writing and checking the Sendables,  I realize Store is actually Sendable - so ignore my comment, no sharing possible, I think.

Sources/AST/RegoValue+PatchMerge.swift
philipaconrad: Okay, so after looking over the implementation again, I think it's correct.

The recursion is confusing here because it's following the same API pattern as `patch(with:at:[String])` and `patch(with:at:ArraySlice<String>)`, where the second function using the `ArraySlice` is the recursive implementation the first function uses to implement everything.

The recursive function is self-sufficient on its own, the `[String]` function is essentially a convenience wrapper around the recursive func.

I've adjusted some variable names and comments to hopefully make this a bit clearer.

Tests/ASTTests/RemovingTests.swift
philipaconrad: I'll add some tests for that presently. I've looked at how OPA does deletions, and I may expand things to handle indexing in arrays too while we're here.

Sources/AST/RegoValue+PatchMerge.swift
koponen: The confusion stems from the fact that the first "if path.isEmpty" is hardcoded to return an object. That doesn't make sense for anything but the root, so I think you could move that to the caller of this function, executed before ever invoking this.

koponen (dismissed):
LGTM in overall. That clarification to the base case of recursion is a minor nit.

Sources/AST/RegoValue+PatchMerge.swift
philipaconrad: Ah! I see what you're getting at. Since visibility of these symbols is fairly limited, there's little risk of the `ArraySlice` variant getting called anywhere unexpected, so I've gone ahead and hoisted the check.

philipaconrad:
ℹ️ Updated to catch up with `main` via a rebase in the Github Web UI...

PR #153, opened 2026-06-02
perf: Replace ==/!- .undefined comparisons with case .undefined pattern matching

koponen:
This allows the compiler better inline than with the synthesized (non)equal functions. Applied to isDefined1, isUndefined1, resultSetAdd1 compact opcodes in executeBlock, execAssignVarOnce, execCall, and execCallDynamic.

Constant improvements throughout the benchmarks, from 3-4% up to 8% with array iter large. No regressions.

### DIFF
--- a/Sources/AST/RegoValue.swift
+++ b/Sources/AST/RegoValue.swift
@@ -107,3 +107,41 @@ extension RegoValue {
         }
     }
 }
+
+// MARK: - Equatable
+
+extension RegoValue {
+    /// Explicit Equatable implementation, marked @inline(__always) so every intra-module call site
+    /// gets a direct discriminant tag comparison rather than a non-inlined
+    /// `__derived_enum_equals` dispatch. Semantically identical to the synthesized
+    /// conformance and consistent with the synthesized Hashable implementation.
+    @inline(__always)
+    public static func == (lhs: RegoValue, rhs: RegoValue) -> Bool {
+        switch lhs {
+        case .undefined:
+            if case .undefined = rhs { return true }
+            return false
+        case .null:
+            if case .null = rhs { return true }
+            return false
+        case .boolean(let a):
+            if case .boolean(let b) = rhs { return a == b }
+            return false
+        case .number(let a):
+            if case .number(let b) = rhs { return a == b }
+            return false
+        case .string(let a):
+            if case .string(let b) = rhs { return a == b }
+            return false
+        case .array(let a):
+            if case .array(let b) = rhs { return a == b }
+            return false
+        case .object(let a):
+            if case .object(let b) = rhs { return a == b }
+            return false
+        case .set(let a):
+            if case .set(let b) = rhs { return a == b }
+            return false
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


philipaconrad:
This change is a bit trickier-- it drops usage of `context.resolveLocal` in a few places, so I'm spending a few minutes making sure we're still correct everywhere. I'd have preferred that we keep `context.resolveLocal` instead of direct indexing, just for consistency. Assuming it is getting inlined appropriately, the performance diff should be negligible. 🤔

Sources/AST/RegoValue.swift
```diff
@@ -107,3 +107,41 @@ extension RegoValue {
         }
     }
 }
+
+// MARK: - Equatable
+
+extension RegoValue {
+    /// Explicit Equatable implementation, marked @inline(__always) so every call site
```
philipaconrad:
```suggestion
    /// Explicit Equatable implementation, marked @inline(__always) so every intra-module call site
```

Sources/Rego/VM+Instructions.swift
```diff
@@ -49,16 +47,15 @@ extension VM {
         let (sourceOp, _) = context.decodeOperand(from: payload, at: start + 4)
         let source = context.resolveUnchecked(sourceOp)
 
-        let currentValue = context.resolveLocal(idx: target)
-        guard currentValue == .undefined else {
-            // Repeated assignments can only be of the same value, otherwise error
-            if currentValue != source {
+        switch context.locals[target] {
```
philipaconrad, line 53: I had to go and measure locally to see why we were specializing here instead of just using `context.resolveLocal`.

It turns out there is a small-but-significant difference in instruction overheads from using the function instead of the direct indexing. The direct approach basically gets to skip most of the cold branches in `resolveLocal`, which is wild.

Sources/Rego/VM.swift
```diff
@@ -152,19 +152,25 @@ extension VM {
             switch opcodeRaw {
             // ── Compact opcodes (rawValues 35–41): operand encoded in `length` field ─────────
             case Int(Opcode.isDefined1.rawValue):
-                if context.resolveLocal(idx: Local(UInt32(length))) == .undefined { return .undefined }
-                continue
+                switch context.locals[Local(UInt32(length))] {
```
philipaconrad, line 158: And a similar story here in the VM too-- there's a small reduction in instruction / call count.

philipaconrad (dismissed):
I was suspicious of dropping `context.resolveLocal` until I measured the instruction count diffs myself.

If you use `@inlinable` so that `RegoValue.==` inlines across module boundaries, it bloats the instruction counts massively with cold branches that are never taken at the call sites.

The direct-indexing approach here + separate branches for `.undefined` look uglier, but probably will be genuinely a little faster on the hot paths.

@koponen I think this is good to merge-- I'd appreciate a comment or something noting the rationale for the direct indexing, but other than that, I think this changeset is good to ship.

philipaconrad (approved):
Looks good, and thanks for adding the comments!

PR #154, opened 2026-06-02
perf: Flatten VMContext memo cache from a stack to a direct field

koponen:
This eliminates the `isEmpty` guard and array index arithmetic on every 2-arg function call in `execCall`'s fast path, where memo lookup is on the hot path. `memoStack` is used only when `with` opcodes are active(rare).

It eliminates one heap allocation per policy evaluation across the entire benchmark suite, with a consistent 1–3% instruction reduction in most benchmarks.

### DIFF
--- a/Sources/Rego/VM.swift
+++ b/Sources/Rego/VM.swift
@@ -313,7 +313,12 @@ internal final class VMContext {
 
     var maxCallDepth: Int = 16_384
     var callDepth: Int = 0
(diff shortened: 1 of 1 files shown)


philipaconrad (dismissed):
This looks good. Let's merge #153 first, so that the diff is just the relevant commit here.

philipaconrad (approved):

PR #155, opened 2026-06-03
perf: RegoNumber intValue fast path for common int case.

koponen:
Split `RegoNumber.intValue` into an `@inline(__always)` fast path for the common `.int` storage case and a separate `@inline(never)` slow path for `.decimal`.

This delivers 20+% time improvement on array-heavy benchmarks with no instruction regression. Other benchmarks are flat.


philipaconrad (approved):
Looks good! The hot path is hotter now. 🔥

PR #156, opened 2026-06-03
perf: Add a number-number fast path to RegoValue.compare.

koponen:
Add a number-number fast path to RegoValue.compare that calls NSDecimalCompare exactly once, eliminating the double evaluation (lhs < rhs then lhs > rhs).

This eliminates 3% of the allocations from the NumericLiterals benchmark.


philipaconrad (approved):
Looks good to me. 🙂

PR #157, opened 2026-06-03
perf: Eliminate per-call RandomNumberGenerator allocation in builtin dispatch

koponen:
Pre-allocate random generator once per VMContext (evaluation) rather than constructing a new one on every builtin call.

Pre-allocating the RandomNumberGenerator eliminates 99% of allocations on array iteration benchmarks and reduces instruction counts by 18–23% on builtin-heavy workloads. Pure user-function policies pay zero overhead thanks to a static sentinel that avoids both the allocation and any Optional nil-check cost.


philipaconrad (approved):
This is clever, and it updates Swift OPA to match my mental model of where RNGs should be allocated-- one RNG per VM context. Nice find! 🙂

PR #158, opened 2026-06-03
perf: Eliminate NSNumber bridge allocation in Decimal.safeInt64Value

koponen:
Skip the NSNumber bridge for non-integer decimals. NSDecimalCompact strips trailing zeros from the significand; if the exponent is still negative after compaction the value is not an integer and we return nil immediately without allocating.

For numeric literals benchmark, this delivers 23% reduction in mallocs and up to 13% wall clock/instruction count improvement. A minor regression with ArrayIter medium for an unknown reason.


philipaconrad (approved):
Looks good. Thanks for adding the fallback for non-Apple platforms!

PR #159, opened 2026-06-04
perf: Replace arr.enumerated() with manual index counter in execScan

koponen:
Avoids the tuple allocation overhead of `enumerated()` on each array element; uses an overflow-safe `&+=` increment instead.

This provides a minor improvement in array iter benchmarks.


philipaconrad (approved):
Looks good!

I had to look up [overflow operators](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/advancedoperators/#Overflow-Operators) to learn about how `&+` works-- unlike many other programming languages, Swift actually errors on numeric overflow instead of allowing it to happen silently by default!

The `&`-prefixed versions of `*`, `+`, and `-` allow the usual over/under-flow behaviors we expect for integers, and make sense to use in this context for efficiency.

PR #160, opened 2026-06-04
perf: Replace Function.path [String] with pre-joined path String

koponen:
Store path as a dot-joined `String` computed once at Function init time, eliminating the `[String]` heap allocation per function and the `path.joined()` call on every `execCallDynamic` comparison.

This delivers 1–3% instruction reduction on dynamic call benchmarks.


philipaconrad (approved):
This is a good, straightforward micro-optimization. It takes work that we were doing frequently at usage time, and moves it to init-time.

PR #161, opened 2026-06-04
perf: Pre-resolve builtins by string table index at query prepare time

koponen:
Replace the per-call `[String: Builtin]` hash lookup in `execCall` with an array subscript keyed by string table index. The index is built once in `PreparedQuery` where policies and the builtin registry first meet, and passed through to VM on each evaluate call.

This results in 2–7% fewer instructions and 5–20% wall clock improvement across all builtin-heavy benchmarks. No regressions.

### DIFF
--- a/Sources/Rego/Engine.swift
+++ b/Sources/Rego/Engine.swift
@@ -102,6 +102,8 @@ extension OPA.Engine {
         let evaluator: any Evaluator
         let store: any OPA.Store
         let builtinRegistry: BuiltinRegistry
+        /// Builtin functions pre-resolved by string table index for each policy.
+        let builtins: [[Builtin?]]
 
         /// Returns the result of evaluating the prepared query against the given input.
         ///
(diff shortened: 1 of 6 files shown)


Sources/Rego/Engine.swift
```diff
@@ -99,9 +99,11 @@ extension OPA.Engine {
     /// PreparedQuery can be re-used for multiple evaluations against different inputs.
     public struct PreparedQuery: Sendable {
         let query: String
-        let evaluator: any Evaluator
+        let evaluator: IREvaluator
```
philipaconrad: Was the concretization of the evaluator type here intentional?

I saw we ripped out the `Evaluator` protocol later on, but could we not have extended or changed the protocol instead of removing it? 🤔 

My expectation before reading the code was that we would not necessarily need to rip out a protocol in order to support the resolved builtins table.

koponen: Fair, that is unrelated to the optimization itself. Removed.

philipaconrad (approved):

PR #162, opened 2026-06-05
perf: Replace two-step check-then-insert with updateValue in execObjectInsertOnce

koponen:
Reduces from 2 dictionary hash+probe operations to 1 for the common case of inserting new keys when building a literal object.

The impact is minor, but measurable: ~40 instructions per key inserted when building literal objects.


philipaconrad (dismissed):
Looks like a straightforward micro-optimization to me! 😄

philipaconrad (approved):

PR #163, opened 2026-06-05
 perf: Check integrality before range bounds in safeInt64Value

koponen:
Move the exponent < 0 integrality check before the range guard. For non-integer decimals, this returns nil immediately without paying the two NSDecimalCompare calls in `self >= int64Min && self <= int64Max`. This makes the fraction exit path cheaper, while the rest remain the same.

This results in ~25% allocation reduction and 13% wall clock improvement in numeric literal test.


philipaconrad (dismissed):
Looks like a good improvement to me. Thanks for the PR!

philipaconrad (approved):

PR #165, opened 2026-06-06
Add synchronous query evaluation alongside async

arirubinstein:
Adds PreparedQuery.evaluateSync(...), a sync-callable evaluation path with no async overhead, sharing a single VM body with the existing async evaluate().

- Split builtins into sync (Builtin, all ~92 defaults) and async (AsyncBuiltin) behind AnyBuiltin; add customSyncBuiltins: engine param (customBuiltins: unchanged for source compatibility).
- Add OPA.SyncStore (InMemoryStore conforms); its async read delegates to the sync one.
- Make the VM core synchronous (native recursion). Sync builtins run inline; async builtins are bridged by blocking, and such policies run off the cooperative pool on a dedicated thread to avoid deadlock.
- Guard native-recursion stack overflow with a per-call stack-headroom check that throws the graceful "max call depth exceeded" error on any thread.
- Gate evaluateSync on a prepare-time hasAsyncBuiltins flag; throw syncEvaluationUnsupported when async work is required.

Faster than the old async-only engine on both paths (rbac bench: 0.234s sync / 0.232s async vs 0.250s). Full unit suite and ComplianceSuite parity preserved.

### DIFF
--- a/Sources/Rego/Store.swift
+++ b/Sources/Rego/Store.swift
@@ -19,6 +19,13 @@ extension OPA {
         mutating func remove(at: StoreKeyPath) async throws
     }
 
+    /// An ``OPA/Store`` that additionally supports synchronous reads, enabling `evaluateSync`.
+    public protocol SyncStore: Store {
+        /// Synchronously returns the value at the specified key path.
+        /// - Throws: an error if the path is not found in the store.
+        func readSync(from: StoreKeyPath) throws -> AST.RegoValue
+    }
+
     /// InMemoryStore is an in-memory implementation of ``OPA/Store``
     public struct InMemoryStore {
         private var data: AST.RegoValue = AST.RegoValue.object([:])
@@ -57,6 +64,21 @@ extension OPA.InMemoryStore: OPA.Store {
     }
 
     public func read(from path: StoreKeyPath) async throws -> AST.RegoValue {
+        return try readSync(from: path)
+    }
+
+    public mutating func write(to path: StoreKeyPath, value: AST.RegoValue) async throws {
+        // TODO this is not achieving our goal of thread safety
+        data = data.patch(with: value, at: path.segments)
+    }
+
+    public mutating func remove(at path: StoreKeyPath) async throws {
+        data = try applyRemove(data, at: path.segments[...], fullPath: path)
+    }
+}
+
+extension OPA.InMemoryStore: OPA.SyncStore {
+    public func readSync(from path: StoreKeyPath) throws -> AST.RegoValue {
         var current: AST.RegoValue = data
         for segment in path.segments {
             switch current {
@@ -76,15 +98,6 @@ extension OPA.InMemoryStore: OPA.Store {
         }
 
         return current
-    }
-
-    public mutating func write(to path: StoreKeyPath, value: AST.RegoValue) async throws {
-        // TODO this is not achieving our goal of thread safety
-        data = data.patch(with: value, at: path.segments)
-    }
-
-    public mutating func remove(at path: StoreKeyPath) async throws {
-        data = try applyRemove(data, at: path.segments[...], fullPath: path)
     }
 }
 
--- a/Sources/Rego/VM.swift
+++ b/Sources/Rego/VM.swift
@@ -7,34 +7,97 @@ import typealias IR.Local
 let localIdxInput = Local(0)
 let localIdxData = Local(1)
 
+/// A value handed off to a worker thread/task while the originating context waits for that work
+/// to finish. The originator never touches the value until the worker completes, so there is no
+/// concurrent access — which is what makes `@unchecked Sendable` sound here. File-scoped `private`
+/// so the escape hatch can't be reused casually elsewhere.
+private struct Handoff<T>: @unchecked Sendable {
+    let value: T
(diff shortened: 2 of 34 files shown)


Sources/Rego/Store.swift
```diff
@@ -57,6 +64,21 @@ extension OPA.InMemoryStore: OPA.Store {
     }
 
     public func read(from path: StoreKeyPath) async throws -> AST.RegoValue {
+        return try readSync(from: path)
+    }
+
+    public mutating func write(to path: StoreKeyPath, value: AST.RegoValue) async throws {
+        // TODO this is not achieving our goal of thread safety
```
arirubinstein, line 71: pre-existing

Sources/Rego/VM.swift
```diff
@@ -4,35 +4,147 @@ import Foundation
...
+}
+
+/// True if the current stack pointer is still above `lowWaterMark` (or the guard is disabled).
+@inline(__always)
+func hasStackHeadroom(_ lowWaterMark: UInt) -> Bool {
+    guard lowWaterMark != 0 else { return true }
+    var probe = 0
+    let sp = withUnsafePointer(to: &probe) { UInt(bitPattern: $0) }
+    return sp > lowWaterMark
+}
```
shomron: I'm not sure it's worth introducing this platform-specific code and pointer arithmetic. Can we just implement a simple call depth limit rather than trying to introspect the stack address space?

arirubinstein: A fixed call-depth limit can't be both safe and usable here. I measured the native recursion depth at which the C stack overflows vs. the depth real policies actually reach:

- Swift runs async code (and the test runner) on the cooperative thread pool, whose threads have small stacks - they overflow at ~40 native recursion levels in debug.
- But legitimate policies also reach ~40 deep: across all 2,067 compliance cases the max was 14, and unit tests peak around 40.

So any single fixed limit would have to be < ~40 to not reject legitimate policies. The obvious workaround (run the VM on a dedicated large-stack thread so a fixed limit is safe) costs ~5x. I prototyped it with a pooled 16mb worker thread, removing all the platform code, but the per-evaluation thread hop is ~7 µs, and a small evaluation is only ~1.6 µs.

The stack probe avoids both problems. By reading the running thread's actual stack bounds it adapts automatically - safe on a 256 KB cooperative stack and on a 16 MB one - with zero per-call overhead.

shomron: I'm not sure why we can't use a safe limit and let the application embedding the Engine override the max depth based on its needs and how it configures the execution context.

arirubinstein: Can do that too, maxCallDepth is now configurable on Engine init (default 32), and the platform-specific stack probe is gone entirely

Sources/Rego/VM.swift
```diff
@@ -7,34 +7,97 @@ import typealias IR.Local
...
+/// set, i.e. on the dedicated thread from `executeSyncOnDedicatedThread`.
+func blockingInvoke(
+    _ builtin: @escaping Builtin,
+    _ ctx: BuiltinContext,
+    _ args: [AST.RegoValue]
+) throws -> AST.RegoValue {
+    let semaphore = DispatchSemaphore(value: 0)
+    let box = BlockingResultBox()
+    let payload = Handoff((builtin, ctx, args))
+    Task.detached {
```
philipaconrad, line 52: [`Task.detached`](https://developer.apple.com/documentation/swift/task#Creating-a-Detached-Task) / [`semaphore.wait()`](https://developer.apple.com/documentation/dispatch/dispatchsemaphore/wait()#Blocking-on-the-Semaphore) sever the cancellation chain, right?

If I understand this correctly, a long-running async builtin will just block until it finishes, and can't be interrupted at all.

It can't be canceled midway by the existing engine, because the engine only runs the cancel check on the synchronous side every 256 instructions. So if an `http.send` builtin is added in the future, it could hang for ages and would always run to completion. 😬

philipaconrad (commented):
Thank you @arirubinstein for putting together this PR! It's been interesting to compare the differences in implementations between this and @koponen's #136 PR, and I deeply appreciate having 2x architectures that I can compare for a large change like this one.

We've so far avoided being opinionated about Swift concurrency, so I'll note that I'm a little reluctant to introduce any flavor of threading or locking into the library just yet. 🤔 
 - I may also have spotted a design issue around cancellation for async builtins-- see the review comments below.

Benchmarking notes:
- The sync paths for both PRs had very similar performance when I ran them through some toy RBAC benchmarks, and were wicked fast. 🙂 
- For async builtins (like a `time.sleep(N)` builtin), the dispatching overheads in this PR's approach caused it to be around 3x slower than Teemu's approach for builtins that returned quickly.
  - I decided to also try an "overlapped eval" case with 300ms delays and a high concurrency level, to see if the dispatching might help in a way that made up for the overheads, and it did not yield too much benefit.

I'm currently leaning strongly towards @koponen's approach, as it has a significantly lower runtime cost for async builtins, and requires no concurrency changes in the library, or any changes in how builtins are written.

arirubinstein:
Thanks for the review!  I think @koponen's approach is fine too - it seems like it has more traction behind it as well. I'm going to close this as I concur.

PR #168, opened 2026-06-23
IREvaluator: Apply SyncSafePatcher to user-supplied IR policies

### LINKED ISSUE #167: IREvaluator skips SyncSafePatcher for user-supplied IR policies, causing requiresAsyncEvaluation for every plan
### Summary

`IREvaluator.init` runs `SyncSafePatcher.patch` on policies derived from bundles, but skips that step for raw policies passed via the
`userPolicies:` parameter. As a result, every user-supplied plan is loaded with `syncSafe == false`, and any caller that uses
`OPA.Engine.init(policies:store:...)` and then invokes the synchronous `PreparedQuery.evaluate` overload unconditionally fails with
`RegoError(.requiresAsyncEvaluation, "plan 'eval' requires async evaluation")` — even when the policy only calls synchronous builtins.

### Repro

The compliance suite reproduces this directly. `runTestSync` in `ComplianceSuite/Sources/RegoCompliance/RegoCompliance.swift` constructs the engine
via `OPA.Engine.init(policies: [tc.base.plan], store: store)`, so it exercises the user-policies code path.

```sh
git checkout main
OPA_COMPLIANCE_TESTS=aggregates make test-compliance
```

aggregates is a fully implemented builtin family (count, sum, min, max, etc.), so all failures should be attributable to engine-level behavior
rather than missing builtins.

Observed
```
✔ Test testCompliance(tc:) with 72 test cases passed after 0.045 seconds.
✘ Suite "Compliance Tests" failed after 0.045 seconds with 70 issues.
✘ Test run with 2 tests in 1 suite failed after 0.045 seconds with 70 issues.
```

Every sync failure has the same shape:
```
✘ aggregates/test-aggregates-0001.json: aggregates/count -> eval
   testFailed(reason: "unexpected evaluation error: RegoError(
     code: .requiresAsyncEvaluation,
     message: \"plan 'eval' requires async evaluation\",
     cause: nil)")
```

Identical failures occur for regex, regexsplit, regexmatch, regexreplace, regexisvalid, regexmatchtemplate, regexfind, and
regexfindallstringsubmatch — basically any compliance category that exercises a fully implemented builtin family.

### Expected

Sync compliance should match async compliance for any plan whose builtins all have sync implementations — currently aggregates (and most other
categories) qualify.

Why bundles work but user policies don't

Sources/Rego/IREvaluator.swift (lines 96–113) for the bundles loop:
```swift
let bytecodePolicy: Bytecode.Policy
do {
    let rawPolicy = try Converter.convert(parsed)
    bytecodePolicy =
        optimizeAsync
        ? SyncSafePatcher.patch(policy: rawPolicy, builtins: builtinRegistry)
        : rawPolicy
} catch { ... }
compiled.append(bytecodePolicy)
```

Same file (lines 119–126) for the user-policies loop:
```swift
for (i, raw) in userPolicies.enumerated() {
    var p = raw
    try p.prepareForExecution()
    for plan in p.plans?.plans ?? [] {
        recordOwner(planName: plan.name, owner: .user(i))
    }
    compiled.append(try Converter.convert(p))   // ← no SyncSafePatcher call
}
```

Without the patcher, the per-plan syncSafe bit defaults to false, and Generated/SyncPeers.swift then throws `requiresAsyncEvaluation` at
`evaluate(...)` time

Unit tests that prepare engines from bundles (`testValidEvaluationsSync` in `Tests/RegoTests/IREvaluatorTests.swift`) pass because the bundles path
already runs the patcher — which is why this regression slipped past the unit suite and only shows up in testComplianceSync.

### Proposed fix

Mirror the bundles-loop patcher invocation in the user-policies loop, gated on the same optimizeAsync flag.

### Environment

- Swift 6.0+, macOS 13+
- Reproduces on commit 7437b1c (Release 0.0.6) and earlier sync-eval-bearing commits (introduced in 8e6e6c5 "rego: Add synchronous evaluation path
via codegen").

DFrenkel:
`IREvaluator.init` ran the sync-safety analyzer on bundle-derived policies but skipped user-supplied raw IR policies, leaving every user-policy plan with `syncSafe = false`. Calling the sync `PreparedQuery.evaluate` on those plans then unconditionally threw `requiresAsyncEvaluation`, even when the plan only called sync builtins.

This was visible in the compliance suite: `runTestSync` constructs the engine with `OPA.Engine.init`, so every `testComplianceSync` case failed with `requires async evaluation`, regardless of which builtins the policy used.
Unit tests that prepare engines from bundles (e.g. `testValidEvaluationsSync`) were unaffected because the bundles path
already runs the patcher.

The fix here just mirrors the bundles-loop patcher invocation in the user-policies loop, gated on the same optimizeAsync flag.

### Definition of done
Compliance tests for fully implemented builtins now pass in both sync and async forms

### How to test
`OPA_COMPLIANCE_TESTS=aggregates make test-compliance` fully succeeds

### Related Resources
Fixes: #167

### DIFF
--- a/Sources/Rego/IREvaluator.swift
+++ b/Sources/Rego/IREvaluator.swift
@@ -122,7 +122,12 @@ internal struct IREvaluator {
             for plan in p.plans?.plans ?? [] {
                 recordOwner(planName: plan.name, owner: .user(i))
             }
-            compiled.append(try Converter.convert(p))
+            let rawPolicy = try Converter.convert(p)
+            compiled.append(
+                optimizeAsync
+                    ? SyncSafePatcher.patch(policy: rawPolicy, builtins: builtinRegistry)
+                    : rawPolicy
+            )
         }
 
         // Ensure all plan names (raw user policies + bundle policies) are unique.


philipaconrad (approved):
The fix looks good, and thank you for catching this.

There's definitely some follow-up work to be done, as noted over in #167. This gets us back to baseline though, which is great!

PR #171, opened 2026-06-23
ci: Allow `diffs` PR check to surface new errors.

### LINKED ISSUE #169: Surface unexpected errors in Compliance test suite in CI
## Description

As noted in #167, we can have substantial breakages happen in the Compliance test suite, and no one will know unless they were running the test suite before and after a change, and counting the number of failures carefully.

In our current CI workflow, we've still got a hack around from the early days: `make test-compliance || true`. This means that failed compliance tests (and there are many still, 500+) don't break the CI run, and the results *can* be inspected if anyone wants to.

The reality however, is that almost no one actually checks the logs unless something breaks. We need to filter expected and unexpected errors apart, and surface the unexpected ones in CI so that problems like #167 don't slip our notice again.

## Proposed solutions

We could filter the compliance test failures against a list of Swift OPA's supported builtins in a script. Any cases where a supported builtin is failing should be surfaced immediately. We can then assume everything else is probably an expected failure and ignore it.

Alternately, if there's an issue where the existing `diff` job isn't detecting failures it should, let's fix that first.

philipaconrad:

This PR reworks the `diffs` PR check to report any new Compliance test errors caused by a PR. The job would report diffs, but could not error when new breakages occurred.

Resolves: #169

### How to test

 - Make a change that breaks things in the Compliance test suite.
 - See if the `diffs` job errors.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -104,8 +104,8 @@ jobs:
       - name: compliance tests (main)
         run: |
           (make test-compliance || true) > results
-          ../../../pr/swift-opa/ComplianceSuite/tools/passed.sh < results > ../../../main-passed
-          ../../../pr/swift-opa/ComplianceSuite/tools/failed.sh < results > ../../../main-failed
+          ../../../main/swift-opa/ComplianceSuite/tools/passed.sh < results > ../../../main-passed
+          ../../../main/swift-opa/ComplianceSuite/tools/failed.sh < results > ../../../main-failed
         working-directory: main/swift-opa/ComplianceSuite
       - name: compliance tests (pr)
         run: |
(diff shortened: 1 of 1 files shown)


philipaconrad:
ℹ️ I'm going to try adding a test commit onto the dev branch that reverts the fix in #168, and we'll see if the new and improved `diffs` job catches it!

EDIT: Looks like the [next CI run](https://github.com/open-policy-agent/swift-opa/actions/runs/28058174761/job/83065713573?pr=171) caught the errors as intended! I'll revert the offending commit, and then we can get this merged in.

philipaconrad:
ℹ️ The failing `Build Swift 6.0.3, Xcode 16.2` PR check is likely coming from [the `macos-latest` label moving from `macos-15` -> `macos-26` here in June](https://github.com/actions/runner-images/issues/14167). I'll address that failure in a separate PR.

PR #172, opened 2026-06-23
builtins: Allow `numbers.range_step` to use Decimal step size.

philipaconrad:

This PR extends the [`numbers.range_step` builtin](https://www.openpolicyagent.org/docs/policy-reference/builtins/numbers#builtin-numbers-numbersrange_step) to support `Decimal` step sizes.

OPA uses bignums from the Go stdlib to support large values for the range and step sizes, but we do not have nice bignum support in Swift at the moment.

While the solution here is not as comprehensive as bignums, it does increase the number of valid cases we can support from OPA, and will fix a broken test case in our Compliance test suite.

### How to test

 - Run `OPA_COMPLIANCE_TESTS='numbersrange' make test-compliance`
 - Observe that it finally passes now without errors! 🙂

### DIFF
--- a/Sources/AST/DecimalExtensions.swift
+++ b/Sources/AST/DecimalExtensions.swift
@@ -74,4 +74,27 @@ extension Decimal {
         guard self >= Self.int64Min && self <= Self.int64Max else { return nil }
         return Int64(truncating: self as NSNumber)
     }
+
+    /// Whether this Decimal represents a whole number (no fractional part), regardless of
+    /// magnitude. Unlike ``safeInt64Value`` this does not require the value to fit in `Int64`,
+    /// so very large integers such as `99999999999999999999` still report `true`.
+    public var isWholeNumber: Bool {
+        guard !self.isNaN && self.isFinite else { return false }
+
+        // exponent >= 0 means the value is already a whole number (significand * 10^exp).
+        guard exponent < 0 else { return true }
+
+        #if canImport(ObjectiveC)
+            // Compaction strips trailing zeros from the significand. If the exponent is still
+            // negative afterwards, the value is fractional.
+            var copy = self
+            NSDecimalCompact(&copy)
+            return copy.exponent >= 0
+        #else
+            var rounded = Decimal()
+            var copy = self
+            NSDecimalRound(&rounded, &copy, 0, .plain)
+            return rounded == self
+        #endif
+    }
 }
--- a/Sources/AST/RegoNumber.swift
+++ b/Sources/AST/RegoNumber.swift
@@ -133,6 +133,27 @@ public struct RegoNumber: Sendable, Hashable {
         }
     }
 
+    /// Whether this number is a whole number (no fractional part), regardless of magnitude.
+    /// Unlike ``int64Value`` this returns true even when the value is too large for `Int64`.
+    public var isInteger: Bool {
+        switch storage {
+        case .int(_):
+            return true
+        case .decimal(let v):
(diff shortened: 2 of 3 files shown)


DFrenkel:
🚀!  I noticed that the same `numbersrangestep/test-numbersrangestep.json: numbersrangestep/big int step` shows as both passed and failed in compliance tests diffs. Do we possibly have another case of sync vs async issue?

Sources/AST/DecimalExtensions.swift
```diff
@@ -74,4 +74,27 @@ extension Decimal {
         guard self >= Self.int64Min && self <= Self.int64Max else { return nil }
         return Int64(truncating: self as NSNumber)
     }
+
+    /// Whether this Decimal represents a whole number (no fractional part), regardless of
+    /// magnitude. Unlike ``safeInt64Value`` this does not require the value to fit in `Int64`,
+    /// so very large integers such as `99999999999999999999` still report `true`.
+    public var isWholeNumber: Bool {
```
DFrenkel, line 81: Should we cover this in `Tests/ASTTests/DecimalExtensionsTests.swift`

Sources/AST/RegoNumber.swift
```diff
@@ -133,6 +133,27 @@ public struct RegoNumber: Sendable, Hashable {
         }
     }
 
+    /// Whether this number is a whole number (no fractional part), regardless of magnitude.
+    /// Unlike ``int64Value`` this returns true even when the value is too large for `Int64`.
+    public var isInteger: Bool {
```
DFrenkel, line 138: We have a series of tests that verify `integerValue` in `Tests/ASTTests/RegoValueTests.swift`. I would suggest adding unit tests there to cover `isInteger` and `isPositive` there

Sources/Rego/Builtins/Numbers.swift
```diff
@@ -76,7 +91,31 @@ extension BuiltinFuncs {
...
         }
+        return .array(result)
+    }
+
+    /// Handle `numbers.range_step` when the step is a positive whole number too large for Int64.
+    /// The endpoints still fit in Int64, so the span `|b - a|` is below 2·Int64.max and the
+    /// result holds at most a handful of elements. We iterate in Decimal space (arbitrary
+    /// precision) to avoid Int64 overflow while remaining correct for steps that are only
+    /// slightly larger than Int64.max.
+    private static func rangeWithDecimalStep(intA: Int64, intB: Int64, step: Decimal) -> RegoValue {
```
DFrenkel, line 102: Should we extend `rangeStepTests` in `Tests/RegoTests/BuiltinTests/NumbersTests.swift` with validations for of these big steps or do you think compliance tests are sufficient?

philipaconrad:
> I noticed that the same `numbersrangestep/test-numbersrangestep.json`: `numbersrangestep/big int` step shows as both passed and failed in compliance tests diffs. Do we possibly have another case of sync vs async issue?

Ah! So the red `-` diff under the "Failed" category indicates it's no longer part of the failing set from `main`. The green `+` diff under the "Passed" category indicates we've added that compliance test to the passing set on the PR.

No sync/async issues, just a confusingly-worded diff summary. I'll see about improving the readability of that message in a future PR. 😅

Sources/AST/DecimalExtensions.swift
philipaconrad: I think we could add some test for that. 🤔 I'll work that into my next round of changes for the PR.

PR #178, opened 2026-07-02
Enable using the Crypto 5.0 beta releases

Lukasa:

This patch changes the Package.swift to allow users who want access to preview APIs
from swift-crypto to set a build flag that includes the 5.0 beta release series.

### How to test

Build and test with and without the flag.


philipaconrad (approved):
This looks good. Will merge shortly. 🙂

PR #179, opened 2026-07-02
ci: Add capabilities.json to release artifacts.

philipaconrad:

We should have the `capabilities.json` file included for each release. This PR finally makes that happen.

### How to test

 - We'll have to cut a test release to see.


srenatus:
That’s a great idea! We could do that in other places.

PR #183, opened 2026-07-06
chore: Upgrade swift-tools-version from 6.0 -> 6.1

philipaconrad:

This PR upgrades our [`swift-tools-version`](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/settingswifttoolsversion/) settings to indicate Swift 6.1 is the minimum supported toolchain version for building the project. 

This change gives the project access to [package traits](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/), which will allow us to separate out optional dependencies better.

Also included is a README update, spelling out our "latest - 2" major Swift version support policy going forward.

### How to test

 - ~~Our 6.0.3 builds should fail. I'll remove them after making sure they explode as-intended.~~
 - If our CI builds properly across 6.1 - 6.3 targets on both Linux and MacOS build targets, we should be good.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -15,8 +15,9 @@ permissions:
 
 jobs:
   ci:
-    name: Build Swift ${{ matrix.swift }}${{ matrix.xcode && format(', Xcode {0}', matrix.xcode) || '' }} (${{ matrix.os }})
+    name: Build Swift ${{ matrix.swift }}${{ matrix.xcode && format(', Xcode {0}', matrix.xcode) || '' }} (${{ matrix.container || matrix.os }})
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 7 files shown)


philipaconrad:
ℹ️ I accidentally clicked on the Copilot reviewer instead of Dmitry. 🤦 Ugh.

I've also confirmed that the tools version changes do break the older builds as anticipated. I'll push up a change that fixes our CI workflows too while I'm here.

-----

It turns out that the [swift-actions/setup-swift](https://github.com/swift-actions/setup-swift) repo hasn't been updated in 8+ months, and thus doesn't support Swift 6.3 for some reason. That breaks the linux parts of our CI matrix, so I'm going to look into alternatives, like switching the linux jobs to using the official [Swift docker containers](https://hub.docker.com/_/swift/).

PR #187, opened 2026-07-08
feat(builtins): implement object.filter, remove, subset

### LINKED ISSUE #186: builtins: implement object filter, remove, subset builtins
Implement the [object](https://www.openpolicyagent.org/docs/policy-reference/builtins/object) builtins:
- [x] [`object.remove`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-objectremove)
- [x] [`object.subset`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-objectsubset)
- [x] [`object.filter`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-objectfilter)

### Expected Outcome
* Object builtins can be used in policy evaluations
* All relevant compliance tests pass: 
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectfilter
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectfilteridempotent
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectfilternonstringkey
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectremove
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectremoveidempotent
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/objectremovenonstringkey
  * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/subset

DFrenkel:
Add the remaining `object.*` builtins, matching OPA Go semantics.

`filter`/`remove` accept the keys argument as an array, set, or object (including non-string keys), normalized into a single key set shared by both. object.subset acts on the operand type pair: object/object, set/set, array/array, and the array/set-as-subset case only. Other combinations (e.g. object/set) cause evaluation errors. Subsets are checked recursively for nested objects/sets/arrays, and arrays match as contiguous, in-order slices, not arbitrary subsequences.

### Definition of done
All remaining `object|subset` compliance tests pass.

### How to test
`make fmt lint test`
`OPA_COMPLIANCE_TESTS=".*(object|subset).*" make test-compliance`

### Related Resources
Fixes #186

### DIFF
--- a/Sources/Rego/Builtins/Objects.swift
+++ b/Sources/Rego/Builtins/Objects.swift
@@ -131,4 +131,202 @@ extension BuiltinFuncs {
 
         return .object(result)
     }
+
+    // filter returns a new object with only the keys of the input object that are
+    // present in the keys argument. keys may be an array, set, or object.
+    // args
+    // object (object[any: any]) - object to filter keys
+    // keys (any<array[any], object[any: any], set[any]>) - keys to keep in object
+    // returns: filtered (any) - remaining data from object with only keys specified in keys
+    static func objectFilter(ctx: BuiltinContext, args: [AST.RegoValue]) throws -> AST.RegoValue {
+        guard args.count == 2 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        guard case .object(let object) = args[0] else {
+            throw BuiltinError.argumentTypeMismatch(arg: "object", got: args[0].typeName, want: "object")
+        }
+
+        let keys = try keysArgumentAsSet(args[1], arg: "keys")
+
+        // Result holds at most one entry per key that also exists in
+        // the object.
+        var result: [AST.RegoValue: AST.RegoValue] = [:]
+        result.reserveCapacity(Swift.min(keys.count, object.count))
+        for key in keys {
+            if let value = object[key] {
+                result[key] = value
+            }
+        }
+
+        return .object(result)
+    }
+
+    // remove returns a new object with the specified keys removed from the input object.
+    // keys may be an array, set, or object.
+    // args
+    // object (object[any: any]) - object to remove keys from
+    // keys (any<array[any], object[any: any], set[any]>) - keys to remove from object
+    // returns: output (any) - result of removing the specified keys from object
+    static func objectRemove(ctx: BuiltinContext, args: [AST.RegoValue]) throws -> AST.RegoValue {
+        guard args.count == 2 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        guard case .object(let object) = args[0] else {
+            throw BuiltinError.argumentTypeMismatch(arg: "object", got: args[0].typeName, want: "object")
+        }
+
+        let keysToRemove = try keysArgumentAsSet(args[1], arg: "keys")
+
+        guard !keysToRemove.isEmpty else {
+            return .object(object)
+        }
+
+        var result: [AST.RegoValue: AST.RegoValue] = [:]
+        result.reserveCapacity(Swift.max(0, object.count - keysToRemove.count))
+        for (key, value) in object where !keysToRemove.contains(key) {
+            result[key] = value
+        }
+
+        return .object(result)
+    }
+
+    // subset determines if the sub value is a subset of the super value.
+    // args
+    // super (any<object[any: any], set[any], array[any]>) - object, set, or array to check against
+    // sub (any<object[any: any], set[any], array[any]>) - object, set, or array to check
+    // returns: result (boolean) - true if sub is a subset of super
+    static func objectSubset(ctx: BuiltinContext, args: [AST.RegoValue]) throws -> AST.RegoValue {
+        guard args.count == 2 else {
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        return .boolean(try isSubset(super: args[0], sub: args[1]))
+    }
+
+    // MARK: - Private helpers
+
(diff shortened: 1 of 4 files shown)


Sources/Rego/Builtins/Objects.swift
```diff
@@ -131,4 +131,181 @@ extension BuiltinFuncs {
...
+        guard superArr.count >= subArr.count else {
+            return false
+        }
+
+        for start in 0...(superArr.count - subArr.count) {
+            var matched = true
+            for offset in 0..<subArr.count where superArr[start + offset] != subArr[offset] {
+                matched = false
+                break
+            }
```
philipaconrad:
```suggestion
        guard let first = subArr.first else {
            return true
        }
        guard superArr.count >= subArr.count else {
            return false
        }

        // We use a filtered for loop to skip starting indices that don't
        // match the start of the sub-array's sequence.
        for start in 0...(superArr.count - subArr.count) where superArr[start] == first {
            // Match the rest of the sequence normally.
            var matched = true
            for offset in 1..<subArr.count where superArr[start + offset] != subArr[offset] {
                matched = false
                break
            }
```
**[suggestion,performance]**: I benchmarked with this change locally, and by matching only the first element from the sub-array on the outer search loop, it avoids running the inner matching loop until there's a match to the start of the sequence.

This yields a small-but-significant constant speedup when the first value in the sub-array is rare. (On the order of ~1.4x, just from skipping the inner loop machinery altogether.)

In the worst-case (and average case), performance degrades to be *exactly* the same as the original naive $O(N^2)$ algorithm.

philipaconrad (commented):
I found this interesting micro-optimization while I was skimming through the code this afternoon. I'll try to have a more thorough review up soon! 😅

Sources/Rego/Builtins/Objects.swift
```diff
@@ -131,4 +131,181 @@ extension BuiltinFuncs {
...
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        guard case .object(let object) = args[0] else {
+            throw BuiltinError.argumentTypeMismatch(arg: "object", got: args[0].typeName, want: "object")
+        }
+
+        let keys = try keysArgumentAsSet(args[1], arg: "keys")
+
+        var result: [AST.RegoValue: AST.RegoValue] = [:]
```
philipaconrad:
```suggestion
    // Result holds at most one entry per key that also exists in
    // the object.
    var result: [AST.RegoValue: AST.RegoValue] = [:]
    result.reserveCapacity(min(keys.count, object.count))
```
**[note]**: We could prealloc here with something like `result.reserveCapacity(min(keys.count, object.count))`, but that might result in objects with more capacity than needed. (Ex: `keys` is empty, or contains keys that match no keys from the `object`.) 🤔

Sources/Rego/Builtins/Objects.swift
```diff
@@ -131,4 +131,181 @@ extension BuiltinFuncs {
...
+            throw BuiltinError.argumentCountMismatch(got: args.count, want: 2)
+        }
+
+        guard case .object(let object) = args[0] else {
+            throw BuiltinError.argumentTypeMismatch(arg: "object", got: args[0].typeName, want: "object")
+        }
+
+        let keysToRemove = try keysArgumentAsSet(args[1], arg: "keys")
+
+        var result: [AST.RegoValue: AST.RegoValue] = [:]
```
philipaconrad, line 186:
```suggestion
    guard !keysToRemove.isEmpty else {
        return .object(object)
    }

    var result: [AST.RegoValue: AST.RegoValue] = [:]
    result.reserveCapacity(Swift.max(0, object.count - keysToRemove.count))
```
**[micro-optimizations]**: Two ideas here:
 - Return a cheap copy if the keys array is empty.
 - Reserve the floor capacity for the result, assuming all keys in `keysToRemove` remove something from the source object.

Of the two, the second idea is probably more marginal.

Sources/Rego/Builtins/Objects.swift
```diff
@@ -131,4 +131,181 @@ extension BuiltinFuncs {
...
+        }
+
+        return true
+    }
+
+    // isSetASubsetOfArray returns true when every member of subset appears somewhere in the
+    // super array, without regard to ordering.
+    private static func isSetASubsetOfArray(super superArr: [AST.RegoValue], sub subSet: Set<AST.RegoValue>) -> Bool {
+        subSet.isSubset(of: Set(superArr))
+    }
```
philipaconrad:
```suggestion
    private static func isSetASubsetOfArray(super superArr: [AST.RegoValue], sub subSet: Set<AST.RegoValue>) -> Bool {
      // Mirrors OPA's arraySetSubset implementation.
      var unmatched = subSet.count
      for element in superArr {
          if subSet.contains(element) {
              unmatched -= 1
          }
          if unmatched == 0 {
              return true
          }
      }
      return false
  }
```
**[optimization]**: Consider directly mirroring OPA's approach in [`arraySetSubset`](https://github.com/open-policy-agent/opa/blob/afb57c7e1346b9f51bef6792101b96e5f3b5d8c4/v1/topdown/subset.go#L194-L211), since I think it will be much faster / fewer allocs.

The OPA implementation also does no recursion whatsoever. 🤔

philipaconrad (dismissed):
Thanks for the PR, @DFrenkel! Overall, this looks solid. 😄 

Some small comments and optimization notes below (the `arraySetSubset` one is particularly neat, I think):

philipaconrad (approved):

PR #193, opened 2026-07-25
feat(builtins): implement json.filter, json.patch, json.remove

### LINKED ISSUE #192: builtins: implement path-based JSON object builtins (json.filter, json.patch, json.remove)
  Implement the [object](https://www.openpolicyagent.org/docs/policy-reference/builtins/object) JSON builtins:
  - [x] [`json.filter`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-jsonfilter)
  - [x] [`json.patch`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-jsonpatch)
  - [x] [`json.remove`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-jsonremove)

  ### Expected Outcome
  * JSON Object builtins can be used in policy evaluations
  * All relevant compliance tests pass:
    * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/jsonfilter
    * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/jsonfilteridempotent
    * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/jsonpatch
    * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/jsonremove
    * https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1/jsonremoveidempotent

DFrenkel:
Add the object-category JSON builtins (`json.filter`, `json.patch`, `json.remove`), matching OPA Go semantics. They address nested locations by JSON-Pointer path (RFC6901: a `/`-separated string with `~0`/`~1` unescaping, or an array of segments) rather than operating on JSON text.

* `filter`/`remove` take a set or array of paths, collapse them into a single tree, and walk the input once to keep or drop the named locations; a shorter path subsumes a longer one (`"a"` covers `"a/b"`). `patch` applies RFC6902 operations (add/remove/replace/move/copy/test) over objects, arrays, sets, and scalars; a failed operation (bad path, failed `test`, unknown op) becomes undefined under non-strict evaluation.

* `patch` is implemented directly on the immutable `RegoValue` tree rather than porting OPA's mutable `edittree`. edittree avoids re-copying the document as many edits are applied in place. Because `RegoValue` is an immutable value type (copy-on-write), the same RFC6902 behavior comes from plain recursive get/insert/remove functions that return new values. A patch applies its operations to one document and returns the result, so there's no repeated in-place editing for edittree to speed up, and copy-on-write keeps the intermediate copies cheap. It passes the full `json-patch-tests` conformance set this way.

### Definition of done
All relevant `json(filter|remove|patch)` compliance tests pass *except* for existing known issues

### How to test
`make fmt lint test`
`OPA_COMPLIANCE_TESTS=".*json(filter|remove|patch).*" make test-compliance`

### Related Resources
Fixes #192

### DIFF
--- a/Sources/Rego/Builtins/Json.swift
+++ b/Sources/Rego/Builtins/Json.swift
@@ -1,0 +1,593 @@
+import AST
+import Foundation
+
+// Path-based JSON builtins from the "object" category: json.filter, json.remove and
+// json.patch. (json.marshal / json.unmarshal / json.is_valid live in Encoding.swift.)
+//
+// None of these functions parse JSON text. They operate on the already-decoded
+// AST.RegoValue tree. The only "JSON" aspect is the JSON-Pointer-style path syntax
+// (RFC6901) used to address nested locations, and RFC6902 JSON-Patch for json.patch.
+extension BuiltinFuncs {
+
+    // MARK: - json.filter
+
+    // filter returns a new object containing only the paths of the input object named
+    // in the second argument. Paths that don't exist in the object are ignored. A path
+    // that names an interior node keeps that entire subtree.
+    //
+    // e.g. json.filter({"a": {"b": 1, "c": 2}}, {"a/b"}) => {"a": {"b": 1}}
+    //
+    // args
+    // object (object[any: any]) - object to filter
+    // paths (any<array[any], set[any]>) - JSON string paths, or arrays of path segments
+    // returns: filtered (any) - remaining data from object with only the paths specified
+    static func jsonFilter(ctx: BuiltinContext, args: [AST.RegoValue]) throws -> AST.RegoValue {
+        guard args.count == 2 else {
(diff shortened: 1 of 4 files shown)


philipaconrad (approved):
This looks like a correct port of the `json.filter`, `json.patch`, and `json.remove` builtins.

I'm not sure that I agree with all of the rationale around why `edittree`-style structures would be less effective here, but I do agree that porting that sort of bookkeeping structure would have enormously complicated the PR.

If we get complaints around `json.patch`'s performance in the future, we can investigate and see if alternative data structures like the `edittree` design would be worth trying out.

For now though, I think this is good to ship. Thanks @DFrenkel!

